PR #2, opened 2021-01-25
Parse HTTP requests from browser with TCPServer

adrianna-chang-shopify:
```
Challenge: Modify your Ruby program so that it can accept HTTP requests and print out (on the console) 
the request method, request target and HTTP version of each request
```

## Key Learnings
- HTTP is a protocol that allows web browsers to talk to web applications
- HTTP is built on top of the TCP protocol (which is built on top of IP). Remember, TCP is application agnostic!
- HTTP communication comes in the form of **messages**
- Messages have the following structure:
     - a start line; then
     - zero or more header fields; then
     - a blank line; and finally
     - an optional message body.
- A `start line` is either a `request-line` (request from browser / client to server) or a `status-line` (response from server to client)
- Request lines follow the following format: begins with a method token, followed by a single space (SP), the request-target, another single space (SP), the protocol version, and ends with CRLF (carriage return line feed - TLDR from Tom, thsi comes from the typewriter days).
- Our program will only parse `GET` requests for now, which are used to fetch a particular path from the app
     - Path = string
     - App has to decide what the string actually means 

## Changes from Challenge 1
- We get the first line from the client, chomp it to get rid of the CRLF
- We split it on the whitespace delimiter to get the three parts of the request line: 
     * a method token which specifies what kind of request is being made;
     * a request target which identifies what path is being requested; and
     * the version number of the HTTP protocol being used.
- We print out the parts of the request line
- We read the rest of the HTTP message (I'm assuming it's all headers)

## Questions for Tom
- I'm assuming this new program only conforms to HTTP protocol - do I need to add some conditional logic to print out the details of the HTTP start line, _only_ if we're receiving an HTTP message? Can an application follow more than one protocol? Is there a more definitive way to define what kind of protocol an application is conforming to?
- Is the `chomp` on the readline necessary? I saw the resource for this method and assumed I needed to use it somewhere 😆 
- Can `GET` requests send a message body? Or do the methods that only fetch resources (get, head, etc.) not have them?
    - If this is possible, can we get the browser to do it with our request? 
- I'm assuming the remainder of the message is the headers... Are these fixed? (ie. would we see these same set of headers across most GET requests?) Are they browser-specific at all?

### DIFF
--- a/challenge_3/http.rb
+++ b/challenge_3/http.rb
@@ -11,28 +11,28 @@ loop do
   puts "Got a new client!"
   
   # Read the request line
-  # Do I need the chomp? Seems to work fine without it
+  # Chomp is not strictly needed given our use of String#split (which discards
+  # the CLRF for us automatically), but it's nice to encapsulate the reading
+  # of this request line into a single line of code that ensures it is cleaned
+  # up appropriately (no surprises later on!)
   request_line = client.readline.chomp
-  request_line_split = request_line.split(' ')
-
   puts "Parsing HTTP request!"
+  method, target, http_version = request_line.split
 
   puts "\n"
   puts "-" * 90
   puts "\n"
 
-  puts "Request method: #{request_line_split[0]}"
-  puts "Request target: #{request_line_split[1]}"
-  puts "HTTP Version: #{request_line_split[2]}"
+  puts "Request method: #{method}"
+  puts "Request target: #{target}"
+  puts "HTTP Version: #{http_version}"
 
-  puts "\n"
-  puts "-" * 90
-  puts "\n"
+  client.close
 
   # Read the rest of the HTTP message
-  # Are these all headers?
-  puts "Getting rest of HTTP message"
-  while !client.eof?
-    puts client.readline
-  end
+  # The rest is all headers
+  # We'll ignore this for now since we just want the request line
+  # while !client.eof?
+  #   puts client.readline
+  # end
 end
--- a/challenge_3/http.rb
+++ b/challenge_3/http.rb
@@ -1,0 +1,38 @@
+require 'socket'
+
+server = TCPServer.new 1234
+
+# "A request is what a browser will send to your application when you visit a
+#   URL like http://localhost:1234/my/awesome/path."
+
+loop do
+  # Accept a client connection
+  client = server.accept
+  puts "Got a new client!"
+  
+  # Read the request line
+  # Do I need the chomp? Seems to work fine without it
+  request_line = client.readline.chomp
+  request_line_split = request_line.split(' ')
+
+  puts "Parsing HTTP request!"
+
+  puts "\n"
+  puts "-" * 90
+  puts "\n"
+
+  puts "Request method: #{request_line_split[0]}"
+  puts "Request target: #{request_line_split[1]}"
+  puts "HTTP Version: #{request_line_split[2]}"
+
+  puts "\n"
+  puts "-" * 90
+  puts "\n"
+
+  # Read the rest of the HTTP message
+  # Are these all headers?
+  puts "Getting rest of HTTP message"
+  while !client.eof?
+    puts client.readline
+  end
+end


challenge_3/http.rb
```diff
@@ -0,0 +1,38 @@
...
+# "A request is what a browser will send to your application when you visit a
+#   URL like http://localhost:1234/my/awesome/path."
+
+loop do
+  # Accept a client connection
+  client = server.accept
+  puts "Got a new client!"
+  
+  # Read the request line
+  # Do I need the chomp? Seems to work fine without it
```
tomstuart: You’re right, you don’t need it, because (I recently learned!) `String#split` has a special case for `' '` which will just [discard the `CRLF`](https://rubydoc.info/stdlib/core/String:split) because it’s trailing whitespace:

> If `pattern` is a single space, `str` is split on whitespace, with leading and trailing whitespace and runs of contiguous whitespace characters ignored.

I think there’s a small amount of value in the `client.readline.chomp` pattern because it neatly encapsulates the idea of “read the next line from the socket” in a single expression without relying on any later code to handle the `CRLF`, but it doesn’t actually matter when you know that string will immediately be fed to `#split`, so you can remove the `#chomp` here if you like.

challenge_3/http.rb
```diff
@@ -0,0 +1,38 @@
...
+  request_line = client.readline.chomp
+  request_line_split = request_line.split(' ')
```
tomstuart: `' '` is (in a roundabout way) the default, so this could just be `request_line.split` if you like.

challenge_3/http.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  puts "Parsing HTTP request!"
+
+  puts "\n"
+  puts "-" * 90
+  puts "\n"
+
+  puts "Request method: #{request_line_split[0]}"
+  puts "Request target: #{request_line_split[1]}"
+  puts "HTTP Version: #{request_line_split[2]}"
```
tomstuart: This works fine, but it’d be nice to assign each part of the request line to a local variable with a meaningful name (e.g. `request_method`) and use that name here instead of indexing into the array. In fact there’s an opportunity to do that earlier: rather than assigning the result array of `request_line.split` to a single variable, you could use [destructuring assignment](https://po-ru.com/2012/09/25/destructuring-assignment-in-ruby) to break it apart into three separate variables at assignment time.

challenge_3/http.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  puts "\n"
+  puts "-" * 90
+  puts "\n"
+
+  # Read the rest of the HTTP message
+  # Are these all headers?
```
tomstuart: Right now, yes. The HTTP message structure is:

* a start line; then
* zero or more header fields; then
* a blank line; and finally
* an optional message body.

The `GET` request sent by your browser when you visit `http://localhost:1234/my/awesome/path` shouldn’t have a message body, so I would expect the rest of the message to be just headers and a blank line.

challenge_3/http.rb
```diff
@@ -0,0 +1,38 @@
...
+  puts "Getting rest of HTTP message"
+  while !client.eof?
```
tomstuart: There’s no harm in reading the rest of the message — it’s interesting to see what’s in there! — but there’s no need to do it at this stage because we’re only interested in the request line. It’s fine to just `client.close` instead if you prefer.

challenge_3/http.rb
```diff
@@ -0,0 +1,38 @@
...
+    puts client.readline
```
tomstuart: FYI this is another place where you could in theory use `#chomp` but in practice it doesn’t matter: strictly speaking you’re calling `puts` (which adds its own newline) with a string that already ends with `"\r\n"` so maybe you need to `#chomp` it first to avoid a double newline? Well, turns out no, because `puts` is [smart enough](https://rubydoc.info/stdlib/core/IO:puts) to only add its own newline if its argument doesn’t already end with one. Thanks Ruby!

tomstuart (approved):
This is excellent, well done! It works great.

I added a few minor remarks in case they’re interesting to you, but I think this is a great solution. 🎉

tomstuart:
> * I'm assuming this new program only conforms to HTTP protocol - do I need to add some conditional logic to print out the details of the HTTP start line, _only_ if we're receiving an HTTP message? Can an application follow more than one protocol? Is there a more definitive way to define what kind of protocol an application is conforming to?

For the purposes of this exercise it’s fine to assume it’s going to receive a correctly-formed HTTP request. This is why the TCP port system exists: port 80 is [assigned to HTTP](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml) and by default it’s reasonable to expect that connections to that port will be sending HTTP requests. (In our case we’re using a higher-numbered port for convenience but we’d want to put it on port 80 outside of an educational setting.)

But yeah, more generally there is of course no guarantee that a random client from somewhere on the internet will be well-behaved! So if this was a production system we’d want to do _something_ defensive to make sure we were gracefully handling connections that tried to speak some protocol other than HTTP, even if just so we can close the socket quickly and start listening for a new client. This hints towards something we’ll see in a later challenge, which is that handling HTTP is actually a bit fiddly and risky so it’s helpful to delegate it to third-party code rather than write it all ourselves. For now we’ll just think about the happy path to keep it simple 😄 

And yes, there’s nothing stopping us from writing code that reads the first few bytes from the socket and tries to make an educated guess about what protocol is being spoken. I imagine this can get arbitrarily complicated depending on how many different protocols we want to be able to detect!

> * Can `GET` requests send a message body? Or do the methods that only fetch resources (get, head, etc.) not have them?

The answer is slightly subtle. There’s nothing in the HTTP/1.1 specification which forbids a `GET` (or `HEAD`) request from including a message body — it’d still be a valid HTTP message — but [section 4.3.1](https://tools.ietf.org/html/rfc7231#section-4.3.1) of RFC 7231 says:

> A payload within a GET request message has no defined semantics; sending a payload body on a GET request might cause some existing implementations to reject the request.

The “no defined semantics” part means that HTTP/1.1 doesn’t assign any meaning to message bodies in `GET` requests, so they’re basically pointless. The “might […] reject the request” warning is a pragmatic acknowledgement that not all HTTP implementations are necessarily tolerant enough to gracefully accept a message body in a `GET` request because it’s so unusual despite being technically legal.
   
We’ll get the browser to send a `POST` request with a message body in challenge 6.

> * I'm assuming the remainder of the message is the headers... Are these fixed? (ie. would we see these same set of headers across most GET requests?) Are they browser-specific at all?

Although they probably wouldn’t change much between subsequent requests from the same browser to the same server, they’re not fixed — there are lots of different headers with different meanings (see [section 5](https://tools.ietf.org/html/rfc7231#section-5) of RFC 7231) and some of them may well vary per request. You can certainly expect to see different headers from different browsers; perhaps the best example is the `User-Agent` header, whose entire purpose is to identify the type of browser to the server.

challenge_3/http.rb
adrianna-chang-shopify: Ah, interesting! That's a great point about encapsulating the reading of the line from the socket neatly in a single expression - if we were to move away from using `split` which automatically handles discarding the `CLRF` for us, something might break and it might not be clear why!

challenge_3/http.rb
adrianna-chang-shopify: Hehe I was indeed very curious about what else was there and figured I'd print out the rest 😆  I'll comment it out for now and switch to `client.close`.

challenge_3/http.rb
tomstuart: Exactly. So it’s perhaps a kind of modularity / compositionality concern: it’s nice to be able to reason locally about each line of code and have confidence that it’s leaving everything in a consistent and unsurprising state, cos then you don’t have to remember any caveats when you come and make changes later.

challenge_3/http.rb
adrianna-chang-shopify: Ooh, thanks Ruby indeed! 😍

PR #3, opened 2021-02-08
Challenge 4 - building an HTTP response for our client

adrianna-chang-shopify:
Challenge 4 - here, we send an HTTP response back to our web client by writing (https://www.rubydoc.info/stdlib/core/IO#write-class_method) the following to it:
```
status line (ending in CRLF)
* (header field) CRLF  <- 0 or more header fields, each ending with a CRLF
CRLF
message body
```

The status line is: `HTTP_VERSION STATUS_CODE OPTIONAL_TEXT_FOR_STATUS CRLF`

We need to tell the browser what type of content it will be receiving so that it can display it properly in the browser. For this challenge, we are just using a plain text message.

### DIFF
--- a/challenge_4/http_response.rb
+++ b/challenge_4/http_response.rb
@@ -1,0 +1,58 @@
+require 'socket'
+
+server = TCPServer.new 1234
+
+# "A request is what a browser will send to your application when you visit a
+#   URL like http://localhost:1234/my/awesome/path."
+
+# More info here: https://tools.ietf.org/html/rfc7231#section-6
+status_codes = {
+  ok: 200
+}
+
+# Accompanying text for status codes
+status_codes_text = {
+  ok: 'OK'
+}
+
+loop do
+  # Accept a client connection
+  client = server.accept
+  puts "Got a new client!"
+  
+  # Read the request line
+  request_line = client.readline.chomp
+  puts "Parsing HTTP request!"
+  method, target, http_version = request_line.split
+
+  puts "Building response for client!"
+  # The status line is a single line of text containing:
+  # - The HTTP protocol version
+  # - The status code
+  # - Optionally, a piece of text to describe the status code
+  #   Question: Can you make this whatever you want, or will the client expect
+  #   the status code to match a given piece of text? When would you ever send
+  #   a "non-typical" piece of text here?
+  status_line = "#{http_version} #{status_codes[:ok]} #{status_codes_text[:ok]}\n"
+
+  # Feedback from Tom on whether my understanding here is correct:
+  # If you have multiple headers, they would each be on separate lines (ie. with a new
+  # line at the end of each header field)
+  # Then, we need ANOTHER new line to separate the set of header fields from the message body (see L55)
+  header_field = "Content-Type: text/plain\n"
+
+  # Build the message body
+  # From Tom in the doc: "The message body of a response message will be displayed by the browser"
+  # This is why it's important we specify the content type, so the browser knows
+  # how to display the information we send in the body
+  message_body = "Request method: #{method}, Request target: #{target}, HTTP Version: #{http_version}"
+
+  # Send response to client
+  client.write(status_line)
+  client.write(header_field)
+  # New line to separate the headers from the message body
+  client.write("\n")
+  client.write(message_body)
+
+  client.close
+end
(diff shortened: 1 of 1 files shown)


challenge_4/http_response.rb
```diff
@@ -0,0 +1,58 @@
...
+  method, target, http_version = request_line.split
+
+  puts "Building response for client!"
+  # The status line is a single line of text containing:
+  # - The HTTP protocol version
+  # - The status code
+  # - Optionally, a piece of text to describe the status code
+  #   Question: Can you make this whatever you want, or will the client expect
+  #   the status code to match a given piece of text? When would you ever send
+  #   a "non-typical" piece of text here?
```
tomstuart, line 37: [Section 3.1.2](https://tools.ietf.org/html/rfc7230#section-3.1.2) of RFC 7230 says:

> The reason-phrase element exists for the sole purpose of providing a
> textual description associated with the numeric status code, mostly
> out of deference to earlier Internet application protocols that were
> more frequently used with interactive text clients.  A client SHOULD
> ignore the reason-phrase content.

Which is another way of saying: it doesn’t really do anything. It’s convenient extra context for any human who happens to be reading the content of a raw HTTP response for some reason, but HTTP client implementations [are expected](https://tools.ietf.org/html/rfc2119#section-3) to completely ignore it.

challenge_4/http_response.rb
```diff
@@ -0,0 +1,58 @@
...
+  status_line = "#{http_version} #{status_codes[:ok]} #{status_codes_text[:ok]}\n"
```
tomstuart: The [spec](https://tools.ietf.org/html/rfc7230#section-3.1.2) says that the status line ends with `CRLF` (`"\r\n"`) rather than just a newline (`"\n"`), so this isn’t strictly correct, but in practice almost all HTTP clients will forgive it.

challenge_4/http_response.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  # Feedback from Tom on whether my understanding here is correct:
+  # If you have multiple headers, they would each be on separate lines (ie. with a new
+  # line at the end of each header field)
```
tomstuart: That’s correct! (Well, [`CRLF` rather than a newline](https://tools.ietf.org/html/rfc7230#section-3). 😄)

challenge_4/http_response.rb
```diff
@@ -0,0 +1,58 @@
...
+  # Then, we need ANOTHER new line to separate the set of header fields from the message body (see L55)
```
tomstuart: Right, the empty line marks the end of the header fields.

challenge_4/http_response.rb
tomstuart: It’s interesting that you’re reusing the client’s `http_version` here! I hadn’t thought of doing that. We are specifically implementing HTTP/1.1 in this exercise so I’d suggest just hardcoding `'HTTP/1.1'` in order to be explicit about which standard we’re implementing, rather than parroting whatever the client happened to send. (Thought experiment: maybe they’re actually an HTTP/1.0 client from the distant past, or an HTTP/5.0 client from the distant future, and will be really confused if we tell them we’re talking the same version of the protocol when we’re not.)

This is an example of [Postel’s law](https://en.wikipedia.org/wiki/Robustness_principle): it’s helpful that we’re liberally accepting any request that’s in vaguely the right format (e.g. it doesn’t actually bother us if the request line’s HTTP version is total garbage) but it makes misunderstandings less likely if we’re conservative about the correctness of the response we send in reply.

tomstuart (approved):
This is great, well done! A couple of small nitpicks for educational purposes only, but this code works so I think you can ship it whenever you’re ready. :shipit:

challenge_4/http_response.rb
adrianna-chang-shopify: Ah yes, thanks for addressing this! I actually had it with the carriage return first, and then tried it without to see what would happen. My browser was indeed forgiving of this (ie. I didn't see a difference), and so I didn't put it back 😆

adrianna-chang-shopify: Ah right, I completely overlooked the fact that those might not be the same thing! TIL about Postel's law too, thanks for sharing! ❤️

PR #4, opened 2021-02-17
Challenge 5- introducing HTML responses

adrianna-chang-shopify:
In this challenge we:
1) Distinguish between different targets (urls), rendering HTML for the path `/show-data` and a plaintext message otherwise.
2) For the `/show-data` target, we render a list of formatted blog posts (title and content). We store our data in memory as an array of hashes. To build our message, we transform each hash into an HTML-escaped, formatted string. We then wrap it in an `<li>` tag, and then wrap all of the <li> elements in a `<ul>.

The use of `CGI#escapeHTML` is important in case any of our data contains HTML tags.

As with the previous challenge, we write the status line, headers and message body to the client.

### DIFF
--- a/challenge_5/html_response.rb
+++ b/challenge_5/html_response.rb
@@ -1,8 +1,9 @@
+### Require dependencies
 require 'socket'
 require 'cgi'
 
-server = TCPServer.new 1234
-
+### Define some constants ###
+#############################
 # More info here: https://tools.ietf.org/html/rfc7231#section-6
 STATUS_CODES = {
   ok: 200
@@ -15,18 +16,22 @@ STATUS_CODES_TEXT = {
 
 HTTP_VERSION = 'HTTP/1.1'
 
-def data
-  [
-    {
-      title: 'My awesome blog!',
-      content: 'my favourite HTML tags are <p> and <script>'
-    },
-    {
-      title: 'Another cool blog!',
-      content: 'my favourite HTML tags are <br> and <hr>'
-    }
-  ]
-end
+BLOG_DATA = 
+[
+  {
+    title: 'My awesome blog!',
+    content: 'my favourite HTML tags are <p> and <script>'
+  },
+  {
+    title: 'Another cool blog!',
+    content: 'my favourite HTML tags are <br> and <hr>'
+  }
+]
+
+CRLF = "\r\n"
+#############################
+
+server = TCPServer.new 1234
 
 loop do
   # Accept a client connection
@@ -39,7 +44,7 @@ loop do
   method, target, http_version = request_line.split
 
   puts "Building response for client!"
-  status_line = "#{HTTP_VERSION} #{STATUS_CODES[:ok]} #{STATUS_CODES_TEXT[:ok]}\r\n"
+  status_line = "#{HTTP_VERSION} #{STATUS_CODES[:ok]} #{STATUS_CODES_TEXT[:ok]}"
 
   # Check request target to determine what to render for message body / headers
   if target == '/show-data'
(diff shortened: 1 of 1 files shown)


adrianna-chang-shopify:
cc @tomstuart ❤️

challenge_5/html_response.rb
```diff
@@ -0,0 +1,69 @@
...
+}
+
+# Accompanying text for status codes
+STATUS_CODES_TEXT = {
+  ok: 'OK'
+}
+
+HTTP_VERSION = 'HTTP/1.1'
+
+def data
```
tomstuart: This could just as well be a `data` variable rather than a method, since it always returns the same value. In fact a variable would be slightly better since you wouldn’t need to recreate the array object on every request!

challenge_5/html_response.rb
```diff
@@ -0,0 +1,69 @@
...
+  client = server.accept
+  puts "Got a new client!"
+  
+  # Read the request line
+  request_line = client.readline.chomp
+  puts "Parsing HTTP request!"
+  method, target, http_version = request_line.split
+
+  puts "Building response for client!"
+  status_line = "#{HTTP_VERSION} #{STATUS_CODES[:ok]} #{STATUS_CODES_TEXT[:ok]}\r\n"
```
tomstuart: There are several `\r\n`-terminated strings in this file. There’s a small hypothetical maintenance risk here: we need to remember to keep including `\r\n` at the end of every such string if we change them or add more in future, so there’s a chance we’ll either get it wrong (e.g. use `\n` or `\n\r` by mistake) or forget to include it entirely.

I can think of a couple of things we could do to mitigate that risk:

* Extract the string `"\r\n"` into a constant with a name like `CRLF` and append that constant to the strings where it’s needed instead of retyping it verbatim. This makes its presence more obvious and essentially eliminates the risk of making a typo.
* Take the `\r\n` out of those strings entirely and write it to the client explicitly as a kind of separator between the lines. That means that instead of…
  ```ruby
  client.write(status_line)
  client.write(header_field)
  client.write("\r\n")
  client.write(message_body)
  ```
  …we would do something like…
  ```ruby
  client.write(status_line)
  client.write(CRLF)
  client.write(header_field)
  client.write(CRLF)
  client.write(CRLF)
  client.write(message_body)
  ```
  …or alternatively…
  ```ruby
  client.write(status_line + CRLF)
  client.write(header_field + CRLF)
  client.write(CRLF)
  client.write(message_body)
  ```
  …which sort of localises the responsibility of knowing that `\r\n` is the line separator in HTTP messages, rather than spreading it throughout the whole program.

challenge_5/html_response.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  # Check request target to determine what to render for message body / headers
+  if target == '/show-data'
+    # The browser knows how to add <html> tags and <head> / <body> tags for us!
+    header_field = "Content-Type: text/html\r\n"
+    message_body = ""
+    message_body << "<ul>"
```
tomstuart, line 54: I actually really like it that you’re separately initialising `message_body` to `""` and then appending `"<ul>"`, rather than just starting with `message_body = "<ul>"`! 👏🏻 

Again there’s some nice modularity to the structure: _first_ we set up a valid (albeit empty) `message_body`, and _then_ we do some work to populate it. This way if we make any changes to the part of the code which adds the `<ul>…</ul>` to the message body (e.g. commenting it out, duplicating it, putting it inside a conditional) then the whole thing keeps working, because we separated out the responsibility of creating the message body to its own line of code rather than combining the two jobs into one.

tomstuart (approved):
Well, this is pretty much perfect, well done! Please ship it when you’re ready. 👍🏻

challenge_5/html_response.rb
adrianna-chang-shopify: This makes a lot of sense! 🙌

PR #5, opened 2021-02-24
Challenge 6: POST requests 🎉

adrianna-chang-shopify:
In this challenge, we introduced forms, post requests, and redirects!
- The main page of the web application renders an HTML form
      - From [the docs](https://html.spec.whatwg.org/#implementing-the-server-side-processing-for-a-form): "For each form control you want submitted, you then have to give a name that will be used to refer to the data in the submission."
- We specify the method (POST), action (what endpoint / URL should the data be submitted to), and enctype (how we encode the data)
- We added a new target, `/create-post`, that will create a new blog post (it accepts `POST` requests only)
- Inside the `/create-post` branch, we handle parsing the request, extracting headers, and then creating a hash for the new post and pushing it into our `BLOG_DATA` structure
- If a client submits a POST request to `/create-post`, we redirect them to the `/show-data` URL where they can view theiir new post! (This means adding a `Location` header indicating the URL to redirect to, as well as using a status code of [303 - See Other](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303))

Other Takeaways:
* Why is it important to know `Content-Length`?
     * The browser sends data over the network
     * All data might come in at once, but we should imagine it "trickling" over the network, since it might _not_ always come all at once (ie. uploading an image might take a while!)
     * When we want to read data off the socket, we can either: 1) Ask it for as much data as it has, or 2) Ask it to give us X amount of data
     * We need to tell the OS to wait to receive all data in the case where it doesn't come right away in the initial TCP connection
     * We also don't want to pick an arbitrary amount (ie. we should know _exactly_ how much data we'll need to read in the message body) because the OS will wait until it has read the exact right amount, or until the client has closed the connection.. Which can be an indeterminate amount of time 😄

### DIFF
--- a/challenge_6/http_post.rb
+++ b/challenge_6/http_post.rb
@@ -1,0 +1,112 @@
+### Require dependencies
+require 'socket'
+require 'cgi'
+require 'uri'
+
+### Define some constants ###
+#############################
+# More info here: https://tools.ietf.org/html/rfc7231#section-6
+STATUS_CODES = {
+  ok: 200,
+  redirect: 303
+}
+
+# Accompanying text for status codes
+STATUS_CODES_TEXT = {
+  ok: 'OK',
+  redirect: 'See Other'
+}
+
+HTTP_VERSION = 'HTTP/1.1'
+
+BLOG_DATA = 
+[
+  {
+    title: 'My awesome blog!',
+    content: 'my favourite HTML tags are <p> and <script>'
+  },
+  {
+    title: 'Another cool blog!',
+    content: 'my favourite HTML tags are <br> and <hr>'
+  }
+]
+
+CRLF = "\r\n"
+#############################
+
+server = TCPServer.new 1234
+
+loop do
+  # Accept a client connection
+  client = server.accept
+  puts "Got a new client!"
+  
+  # Read the request line
+  request_line = client.readline.chomp
+  puts "Parsing HTTP request!"
+  method, target, http_version = request_line.split
+
+  puts "Building response for client!"
+  status_line = "#{HTTP_VERSION} #{STATUS_CODES[:ok]} #{STATUS_CODES_TEXT[:ok]}"
+  header_field = "Content-Type: text/html"
+
+  # Check request target to determine what to send back to client
+  if target == '/show-data'
+    message_body = ""
+    message_body << "<ul>"
+    BLOG_DATA.each do |element|
+      message_body << "<li>"
+      message_body << "<strong>Title: #{CGI.escape_html(element[:title])}</strong>, Content: #{CGI.escape_html(element[:content])}"
+      message_body << "</li>"
+    end
+    message_body << "</ul>"
+  elsif target == '/create-post' && method == 'POST'
+    status_line = "#{HTTP_VERSION} #{STATUS_CODES[:redirect]} #{STATUS_CODES_TEXT[:redirect]}"
+    # Don't need a Content-Type here, we're redirecting!
+    header_field = "Location: /show-data"
+
+    puts "Got a new POST request!"
+    # Not sure if this is how to do this
+    # Let's try reading all the headers in until we get a line that is just CRLF
+    headers = {}
+    line = client.readline
+    while (line = client.readline) != CRLF
+      header_name, header_value = line.chomp.split(": ")
+      headers[header_name] = header_value
+    end
+    content_length = headers["Content-Length"]
+    body = client.read(content_length.to_i)
+    post = Hash[URI.decode_www_form(body)]
+    post.transform_keys! { |key| key.to_sym }
+    BLOG_DATA << post
+  else
+    # Main Page
+    message_body =  ""
+    message_body << "<p><strong>Submit a new Blog Post!</p></strong>"
+    # Method = POST
+    # Encoding type = application/x-www-form-urlencoded (usual encoding system, Ruby has built-in decoder)
+    # Action = /create-post (Seems like this just needs to be a relative target path, but docs use full URL)
+    message_body << "<form method=\"post\" enctype=\"application/x-www-form-urlencoded\" action=\"/create-post\">"
+    message_body << "<p><label>Blog Title: <input name='title'></label></p>"
+    message_body << "<p><label>Content: <textarea name='content'></textarea></label></p>"
+    message_body << "<p><button>Submit post</button></p>"
+    message_body << "</form>"
+  end
+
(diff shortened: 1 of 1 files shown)


challenge_6/http_post.rb
```diff
@@ -0,0 +1,112 @@
...
+require 'socket'
+require 'cgi'
+require 'uri'
+
+### Define some constants ###
+#############################
+# More info here: https://tools.ietf.org/html/rfc7231#section-6
+STATUS_CODES = {
+  ok: 200,
+  redirect: 303
```
tomstuart: Maybe it would avoid future confusion if you called this key `:see_other`, since that would make it match its `STATUS_CODES_TEXT` value in the same way `:ok` does? If it’s set up differently from `:ok` then a future maintainer might assume that the difference exists for an important reason, but at this stage I think it’s kind of arbitrary, so we might as well be consistent as long as it doesn’t cost us anything.

challenge_6/http_post.rb
```diff
@@ -0,0 +1,112 @@
...
+  request_line = client.readline.chomp
+  puts "Parsing HTTP request!"
+  method, target, http_version = request_line.split
+
+  puts "Building response for client!"
+  status_line = "#{HTTP_VERSION} #{STATUS_CODES[:ok]} #{STATUS_CODES_TEXT[:ok]}"
+  header_field = "Content-Type: text/html"
+
+  # Check request target to determine what to send back to client
+  if target == '/show-data'
```
tomstuart: Again, in the spirit of avoiding unintended inconsistency, maybe you could start checking that `method` is `'GET'` here, in the same way as you do for the `/create-post` target? Otherwise the difference between the two conditions might imply that you intend `/show-data` to accept other request methods.

If you do decide to do this, note that another way of saying…

```ruby
if target == '/show-data' && method == 'GET'
  # …
elsif target == '/create-post' && method == 'POST'
  # …
else
  # …
end
```

…is…

```ruby
case [method, target]
when ['GET', '/show-data']
  # …
when ['POST', '/create-post']
  # …
else
  # …
end
```

…if you prefer.

challenge_6/http_post.rb
```diff
@@ -0,0 +1,112 @@
...
+    message_body = ""
+    message_body << "<ul>"
+    BLOG_DATA.each do |element|
+      message_body << "<li>"
+      message_body << "<strong>Title: #{CGI.escape_html(element[:title])}</strong>, Content: #{CGI.escape_html(element[:content])}"
+      message_body << "</li>"
+    end
+    message_body << "</ul>"
+  elsif target == '/create-post' && method == 'POST'
+    status_line = "#{HTTP_VERSION} #{STATUS_CODES[:redirect]} #{STATUS_CODES_TEXT[:redirect]}"
```
tomstuart: To be massively nitpicky: this differs from the previous value of `status_line` only by the key you’re using to fetch the numeric status code and reason phrase. Given that you’ve bothered to set up the whole `STATUS_CODES`/`STATUS_CODES_TEXT` system, you could make use of it by refactoring this code to assign the status code key in each arm of the conditional (or, equivalently, set it to `:ok` by default and reassign it to `:redirect` here) and then assemble `status_line` once at the end as part of sending the response.

challenge_6/http_post.rb
```diff
@@ -0,0 +1,112 @@
...
+    # Don't need a Content-Type here, we're redirecting!
+    header_field = "Location: /show-data"
+
+    puts "Got a new POST request!"
+    # Not sure if this is how to do this
+    # Let's try reading all the headers in until we get a line that is just CRLF
+    headers = {}
+    line = client.readline
+    while (line = client.readline) != CRLF
+      header_name, header_value = line.chomp.split(": ")
```
tomstuart: Thinking about the (imaginary) future: this code would be problematic for any header field whose value contained `': '`. For example, a hypothetical header like `My-JSON-Header: { "name": "Tom" }` would be split into `['My-JSON-Header', '{ "name"', '"Tom" }']` which isn’t what you want.

You could avoid this by using [`String#partition`](https://rubydoc.info/stdlib/core/String:partition) instead, which lets you split only on the first occurrence of the separator (since that’s really what you mean):

```
>> header_name, _, header_value = 'My-JSON-Header: { "name": "Tom" }'.partition(': ')
=> ["My-JSON-Header", ": ", "{ \"name\": \"Tom\" }"]
>> header_name
=> "My-JSON-Header"
>> header_value
=> "{ \"name\": \"Tom\" }"
```

challenge_6/http_post.rb
```diff
@@ -0,0 +1,112 @@
...
+      headers[header_name] = header_value
+    end
+    content_length = headers["Content-Length"]
+    body = client.read(content_length.to_i)
+    post = Hash[URI.decode_www_form(body)]
+    post.transform_keys! { |key| key.to_sym }
```
tomstuart: You could use the [`Symbol#to_proc` shorthand](https://rubydoc.info/stdlib/core/Symbol:to_proc) to write this as `post.transform_keys!(&:to_sym)` if you wanted to, although the block argument is more explicit.

challenge_6/http_post.rb
```diff
@@ -0,0 +1,112 @@
...
+    BLOG_DATA << post
+  else
+    # Main Page
+    message_body =  ""
+    message_body << "<p><strong>Submit a new Blog Post!</p></strong>"
+    # Method = POST
+    # Encoding type = application/x-www-form-urlencoded (usual encoding system, Ruby has built-in decoder)
+    # Action = /create-post (Seems like this just needs to be a relative target path, but docs use full URL)
+    message_body << "<form method=\"post\" enctype=\"application/x-www-form-urlencoded\" action=\"/create-post\">"
```
tomstuart: You can use single quotes for the `<form>` tag attributes here, which would get you out of backslash hell.

tomstuart (approved):
This works great, well done! I’ve left the usual “educational” remarks but this is already totally mergeable. 🎉

challenge_6/http_post.rb
adrianna-chang-shopify: Ooh, the case statement is a pretty neat way to achieve this! 😍

challenge_6/http_post.rb
adrianna-chang-shopify: Ooh, thank you! TIL about `#partition` ❤️

tomstuart:
Love it!

PR #6, opened 2021-03-03
Introduce persistent data store

adrianna-chang-shopify:
In this PR, we introduce a persistent data store so that our blogs will not be lost every time the Ruby program restarts.
This can be done by using [YAML::Store](https://www.rubydoc.info/stdlib/yaml/2.0.0/YAML/Store).

[Psych](https://github.com/ruby/psych) is now the default YAML parser for Ruby, and it can serialize and de-serialize most Ruby objects to and from the YAML format (cool - this means we can actually write Blog structs straight to the YAML and read them as such 🎉).

Other takeaways:
* Why the use of `store.transaction` every time we read from and write to the store?
     * Ensures the property of atomicity is preserved (ie. changes all either succeed or fail, so you don't end up with some changed getting persisted to the file but not others)
     * Ensures optimistic concurrency (avoid data changing as you read it; prevent scenarios where two stores are trying to modify the contents of a YAML at the same time)
* Why does `URI.decode_www_form(body)` produce something that looks like `[["title", "Test"], ["content", "test"]]`?
     * Form names can actually be duplicated (they are not unique)
     * Consequently, we can't decode into a hash because there might be duplicated keys
     * Interesting use case from Tom: Rails actually uses duplicated form fields to handle checkboxes (it inserts a hidden field with the same name, and that is what gets sent if the user filling out the form doesn't tick the checkbox)

### DIFF
--- a/challenge_7/data_store.rb
+++ b/challenge_7/data_store.rb
@@ -1,0 +1,116 @@
+### Require dependencies
+require 'socket'
+require 'cgi'
+require 'uri'
+require 'yaml/store'
+
+### Define some constants ###
+#############################
+# More info here: https://tools.ietf.org/html/rfc7231#section-6
+STATUS_CODES = {
+  ok: 200,
+  see_other: 303
+}
+
+# Accompanying text for status codes
+STATUS_CODES_TEXT = {
+  ok: 'OK',
+  see_other: 'See Other'
+}
+
+HTTP_VERSION = 'HTTP/1.1'
+
+# Struct to define what a Blog looks like
+Blog = Struct.new(:title, :content, keyword_init: true)
+
+CRLF = "\r\n"
+#############################
+
+# Start a TCP Server on port 1234
+server = TCPServer.new 1234
+# Retrieve data from our YAML store
+store = YAML::Store.new(File.expand_path('blogs.yml', __dir__))
+
+loop do
+  # Accept a client connection
+  client = server.accept
+  puts 'Got a new client!'
+
+  # Read the request line
+  request_line = client.readline.chomp
+  puts 'Parsing HTTP request!'
+  method, target, http_version = request_line.split
+
+  puts 'Building response for client!'
+  # Check method type and request target to determine what to send back to client
+  case [method, target]
+  when ['GET', '/show-data']
+    message_body = ''
+    message_body << '<ul>'
+
+    # Transaction in case someone writes to the store in between our reads,
+    # and the data is no longer consistent
+    blog_data = store.transaction { store[:blogs] }
+    blog_data.each do |element|
+      message_body << '<li>'
+      message_body << "<strong>Title: #{CGI.escape_html(element.title)}</strong>, Content: #{CGI.escape_html(element.content)}"
+      message_body << '</li>'
+    end
+    message_body << '</ul>'
+
+    # Prepare response
+    status_code = :ok
+    header_field = 'Content-Type: text/html'
+  when ['POST', '/create-post']
+    puts 'Got a new POST request!'
+    headers = {}
+    line = client.readline
+    while (line = client.readline) != CRLF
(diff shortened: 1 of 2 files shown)


adrianna-chang-shopify:
@tomstuart so I opened this PR and totally forgot to tag you on it 🤦‍♀️

tomstuart:
Oh no! Hello! 😂

challenge_7/blogs.yml
```diff
@@ -0,0 +1,11 @@
+---
```
tomstuart: How do you feel about the trade-offs of committing this YAML file to the repo, versus putting it in the `.gitignore` so that Git doesn’t pay attention to it? What are the downsides either way, e.g. for your experience working on the app, or for someone else cloning the repo for the first time? I don’t think there’s a right answer here, it’s just something to consider. 😄

tomstuart (approved):
This looks good to me! No surprises since we paired on it.

A meta-comment: it’s a little difficult for me to spot exactly what’s changed in each challenge because you start a new file each time. Maybe next time you could commit the new challenge’s file right after you duplicate it from the old one, and then subsequent commits will contain the new changes separately? ❤️

challenge_7/blogs.yml
adrianna-chang-shopify: Oh, this is an interesting one! 😁  So back in challenge 6 I had some "seed" blog data that would get pushed into our in-memory data structure so that I could see data right away without having to create blogs via post requests all the time. I kind of like the idea of keeping seeded data, but having Git update the file and continue to track it as I play around locally is a bit of a pain (and if multiple collaborators were working on this project, we wouldn't want to be mucking with this file all the time I imagine!) 

Maybe the way around this is to not commit this file (and to tell Git to ignore it), and if I care about seeding data I can handle that in the Ruby script itself. I think that makes it more obvious why that data is there / where it came from.

Something else I noticed - right now if you _don't_  have anything in the `blogs.yml` , this blows up:
```ruby
    blog_data = store.transaction { store[:blogs] }
    blog_data.each do |element|
      message_body << "<li>"
      message_body << "<strong>Title: #{CGI.escape_html(element.title)}</strong>, Content: #{CGI.escape_html(element.content)}"
      message_body << "</li>"
    end
```
I noticed this when copying the file over to my `challenge_8` directory, but I didn't copy the `blogs.yml` file so I hit ` #<NoMethodError: undefined method `each' for nil:NilClass>` 😅  
Something I will also address before merging this!

tomstuart: Yeah, interesting! Is there a way we could make it gracefully set up the store if it doesn’t already exist?

challenge_7/data_store.rb
```diff
@@ -31,6 +31,19 @@
...
+end
+
+# Seed some blog data
+# Comment out if you'd like to start from scratch!
+store.transaction do
+  if store[:blogs].empty?
+    store[:blogs] << Blog.new(title: 'My awesome blog!', content: 'my favourite HTML tags are <p> and <script>')
+    store[:blogs] << Blog.new(title: 'Another cool blog!', content: 'my favourite HTML tags are <br> and <hr>')
+  end
+end
```
tomstuart, line 45: This is really cool! 🎉

challenge_7/blogs.yml
adrianna-chang-shopify: I've added https://github.com/adrianna-chang-shopify/learning-project/pull/6/commits/6d4f24ed5458b481c447d059b8dc1823e3f4f5de to:
- set `store[:blogs]` to an empty array if it's nil
- optionally seed some data if `store[:blogs]` is an empty array. This can be commented out though (ie. if a user wanted to start from scratch) and everything should still work 😁

tomstuart: Awesome work, and very user-friendly! :shipit:

PR #7, opened 2021-03-12
Challenge 8

adrianna-chang-shopify:
In this challenge, we introduce a Rack application to handle the TCP / HTTP complexities for us, so we no longer have to speak directly to a TCP server! A Rack app is 
> any Ruby object which responds to the #call message, accepts a single argument called the environment, and returns a three-element array containing a status code, a hash of response headers and a response body.

Neat! Rack has a further specification which defines exactly how the app server talks to our code, and how we should respond. There are many different app servers that conform to [Rack's specification](https://github.com/rack/rack/blob/master/SPEC.rdoc) (Puma is just one of them!). Our web app can actually run with any app server that supports Rack's specification, since we are using this interface, rather than coupling ourselves directly to an implementation (ie. Puma).

### DIFF
--- a/challenge_8/app_server.rb
+++ b/challenge_8/app_server.rb
@@ -1,33 +1,18 @@
 ### Require dependencies
-require 'socket'
 require 'cgi'
 require 'uri'
 require 'yaml/store'
+require 'rack/handler/puma'
 
-### Define some constants ###
-#############################
-# More info here: https://tools.ietf.org/html/rfc7231#section-6
+### Status codes
 STATUS_CODES = {
   ok: 200,
   see_other: 303
 }
 
-# Accompanying text for status codes
-STATUS_CODES_TEXT = {
-  ok: 'OK',
-  see_other: 'See Other'
-}
-
-HTTP_VERSION = 'HTTP/1.1'
-
 # Struct to define what a Blog looks like
 Blog = Struct.new(:title, :content, keyword_init: true)
 
-CRLF = "\r\n"
-#############################
-
-# Start a TCP Server on port 1234
-server = TCPServer.new 1234
 # Retrieve data from our YAML store
 store = YAML::Store.new(File.expand_path('blogs.yml', __dir__))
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1 +1 @@
-blogs.yml
+**/blogs.yml
```
tomstuart: Do we need this? The [`gitignore` docs](https://git-scm.com/docs/gitignore#_pattern_format) say:

> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular `.gitignore` file itself. Otherwise the pattern may also match at any level below the `.gitignore` level.

So I think `blogs.yml` and `**/blogs.yml` have the same meaning. Maybe you prefer the latter?

challenge_8/app_server.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  # Send response to client
+  client.write(status_line + CRLF)
+  client.write(header_field + CRLF)
+  # CRLF to separate the headers from the message body
+  client.write(CRLF)
+  client.write(message_body)
+
+  client.close
+end
```
tomstuart: Thank you for committing the old code first! It makes it much easier to see what you changed for this challenge. :smile:

challenge_8/app_server.rb
```diff
@@ -1,33 +1,18 @@
...
 STATUS_CODES = {
   ok: 200,
   see_other: 303
 }
 
-# Accompanying text for status codes
-STATUS_CODES_TEXT = {
-  ok: 'OK',
-  see_other: 'See Other'
-}
```
tomstuart: 👋🏻  Bye, reason phrases! How come don’t we need to know this any more? Oh right, it’s because [Puma does](https://github.com/puma/puma/blob/v5.2.2/lib/puma/const.rb#L18-L79) 😅

challenge_8/app_server.rb
```diff
@@ -1,33 +1,18 @@
...
-
-HTTP_VERSION = 'HTTP/1.1'
```
tomstuart: 👋🏻 https://github.com/puma/puma/blob/v5.2.2/lib/puma/const.rb#L181

challenge_8/app_server.rb
```diff
@@ -1,33 +1,18 @@
...
-
 # Struct to define what a Blog looks like
 Blog = Struct.new(:title, :content, keyword_init: true)
 
-CRLF = "\r\n"
```
tomstuart: 👋🏻  https://github.com/puma/puma/blob/v5.2.2/lib/puma/const.rb#L165

challenge_8/app_server.rb
```diff
@@ -1,33 +1,18 @@
...
-#############################
-
-# Start a TCP Server on port 1234
-server = TCPServer.new 1234
```
tomstuart: 👋🏻 https://github.com/puma/puma/blob/v5.2.2/lib/puma/binder.rb#L288

Puma [defaults](https://github.com/puma/puma/blob/v5.2.2/lib/puma/configuration.rb#L13) to port 9292, but you can [pass](https://github.com/puma/puma/blob/v5.2.2/lib/rack/handler/puma.rb#L49) `Port: 1234` to `Rack::Handler::Puma.new` if you want to keep the app running on port 1234.

challenge_8/app_server.rb
```diff
@@ -44,86 +29,71 @@
   end
 end
 
-loop do
```
tomstuart: 👋🏻 https://github.com/puma/puma/blob/v5.2.2/lib/puma/server.rb#L322

challenge_8/app_server.rb
```diff
@@ -44,86 +29,71 @@
...
-  # Accept a client connection
-  client = server.accept
```
tomstuart: 👋🏻 https://github.com/puma/puma/blob/v5.2.2/lib/puma/server.rb#L334

Notice that Puma is using the [non-blocking counterpart](https://github.com/ruby/ruby/blob/v3_0_0/ext/socket/lib/socket.rb#L1281-L1308) of `TCPServer#accept` in [combination](https://github.com/puma/puma/blob/v5.2.2/lib/puma/server.rb#L324) with [`IO.select`](https://www.rubydoc.info/stdlib/core/IO%2Eselect). This is a common pattern for dealing with multiple sockets at once: we don’t want to block on handling any one socket in particular because that would starve the other sockets of attention, so we use `IO.select` to block on _all_ the sockets until at least one of them is ready, then use a non-blocking operation (`#accept_nonblock` in this case) on each of the ready sockets that `IO.select` returns.

Another difference is that Puma doesn’t process the client socket synchronously once it’s been accepted — instead it [passes it to a pool of threads](https://github.com/puma/puma/blob/v5.2.2/lib/puma/server.rb#L344) where it gets [pushed into a shared work queue](https://github.com/puma/puma/blob/v5.2.2/lib/puma/thread_pool.rb#L183-L197) and then [picked up](https://github.com/puma/puma/blob/v5.2.2/lib/puma/thread_pool.rb#L137) by the next available thread.

Phew. I’m glad our implementation only had to deal with one server socket on the main thread! 😅

challenge_8/app_server.rb
```diff
@@ -44,86 +29,71 @@
...
-  puts 'Got a new client!'
+app = lambda { |environment|
+  puts 'Rack app got a request!'
+  request_method = environment['REQUEST_METHOD']
+  request_path   = environment['PATH_INFO']
 
-  # Read the request line
-  request_line = client.readline.chomp
-  puts 'Parsing HTTP request!'
-  method, target, http_version = request_line.split
```
tomstuart: 👋🏻 https://github.com/puma/puma/blob/v5.2.2/ext/puma_http11/http11_parser_common.rl#L42

Puma uses the [Ragel](http://www.colm.net/open-source/ragel/) parser generator to [translate](https://github.com/puma/puma/blob/v5.2.2/Rakefile#L22) a description of the [syntax of HTTP](https://github.com/puma/puma/blob/v5.2.2/ext/puma_http11/http11_parser_common.rl) into an [inscrutable C file](https://github.com/puma/puma/blob/v5.2.2/ext/puma_http11/http11_parser.c). Puma, as well as a [competing](https://blog.engineyard.com/ruby-app-server-arena-pt1) Rack application server called [Thin](https://github.com/macournoyer/thin), lifted this entire approach from an earlier (pre-Rack!) Ruby HTTP server called [Mongrel](https://github.com/mongrel/mongrel).

Generating a C parser with Ragel is far more complicated than what we did to parse HTTP in pure Ruby, but unsurprisingly the resulting parser also runs much faster and copes better with unexpected or malicious inputs.

challenge_8/app_server.rb
```diff
@@ -44,86 +29,71 @@
...
+      body << "<strong>Title: #{CGI.escape_html(element.title)}</strong>, Content: #{CGI.escape_html(element.content)}"
+      body << '</li>'
     end
-    message_body << '</ul>'
+    body << '</ul>'
 
-    # Prepare response
-    status_code = :ok
-    header_field = 'Content-Type: text/html'
+    status = STATUS_CODES[:ok]
```
tomstuart: We used to set `status_code` to a friendly symbol (`:ok` here, `:see_other` below) and convert it to a numeric code as late as possible by looking it up in `STATUS_CODES` while building the status line. From a modularity point of view that was actually kind of nice: there was only one line of code which knew about `STATUS_CODES` and everything else could deal in friendly symbols like `:ok` without having to be aware of their numeric equivalents. Do you think it’d be worth doing the same for `status` in this version?

challenge_8/app_server.rb
```diff
@@ -44,86 +29,71 @@
...
+    headers['Content-Type'] = 'text/html'
   when ['POST', '/create-post']
-    puts 'Got a new POST request!'
-    headers = {}
-    line = client.readline
-    while (line = client.readline) != CRLF
-      header_name, _, header_value = line.chomp.partition(': ')
```
tomstuart: 👋🏻 https://github.com/puma/puma/blob/v5.2.2/ext/puma_http11/http11_parser_common.rl#L48

challenge_8/app_server.rb
```diff
@@ -44,86 +29,71 @@
...
+        post[name] = value
+      end
+
+      store.transaction do
+        store[:blogs] << post
+      end
+
+      # Prepare response
+      status = STATUS_CODES[:see_other]
+      headers['Location'] = '/show-data' # NOTE: Don't need a Content-Type here, we're redirecting!
```
tomstuart: The indentation is off here.

challenge_8/app_server.rb
```diff
@@ -44,86 +29,71 @@
...
-  client.write(header_field + CRLF)
-  # CRLF to separate the headers from the message body
-  client.write(CRLF)
-  client.write(message_body)
+  [status, headers, body]
+}
 
-  client.close
-end
+Rack::Handler::Puma.run(app)
```
tomstuart: As Maple discovered, you can [pass](https://github.com/puma/puma/blob/v5.2.2/lib/rack/handler/puma.rb#L33-L36) `Verbose: true` if you want to enable Puma’s logging output.

tomstuart (approved):
This works great, well done! Mostly just footnotes from me this time. 😄

.gitignore
adrianna-chang-shopify: You're right, not needed!

challenge_8/app_server.rb
adrianna-chang-shopify: Thanks Puma! 😄

challenge_8/app_server.rb
adrianna-chang-shopify: Nice, let's do it!!

challenge_8/app_server.rb
adrianna-chang-shopify: This is SO cool, thanks for sharing! Pushing the client socket to a queue to be processed asynchronously by the next available thread reminds me a lot of the work I've been doing with Maintenance Tasks lately, and how ActiveJob handles processing incoming jobs asynchronously by pushing them to a queuing backend to be picked up by a job worker. This is really neat!

challenge_8/app_server.rb
adrianna-chang-shopify: WOW, that's pretty dope! Definitely glad we didn't have to do that ourselves 😂

challenge_8/app_server.rb
adrianna-chang-shopify: Yes, I think that makes a lot of sense!!

adrianna-chang-shopify:
@tomstuart I ran into something interesting!

Screen Shot 2021-03-16 at 10 23 08 AM

This is following the update to Puma v5.2.2, so Maple's changes are in there. It seems to be able to create a new instance of `Rack::CommonLogger` just fine, but then attempts to reference `Rack::CommonLogger::Utils` [here](https://github.com/rack/rack/blob/master/lib/rack/common_logger.rb#L37).

EDIT: I'm a doof 🤦‍♀️  I need to require `rack`, as right now we're only requiring `rack/handlers/puma`.

PR #8, opened 2021-03-17
Challenge 9 - Refactor app to use Rack::Request

adrianna-chang-shopify:
In this challenge, we avoid looking into our Rack application's `environment` hash directly, opting instead to access information about the request through a handy dandy `Rack::Request` object. `Rack::Request` essentially acts as a wrapper to the environment hash, and exposes a number of helpful methods we can use to parse the request and determine our response.

### DIFF
--- a/challenge_9/rack_request.rb
+++ b/challenge_9/rack_request.rb
@@ -1,9 +1,9 @@
 ### Require dependencies
 require 'cgi'
+require 'rack/handler/puma'
+require 'rack'
 require 'uri'
 require 'yaml/store'
-require 'rack'
-require 'rack/handler/puma'
 
 ### Status codes
 STATUS_CODES = {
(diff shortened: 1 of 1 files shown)


challenge_9/rack_request.rb
```diff
@@ -48,25 +52,29 @@
...
+    body << "method is #{request.request_method}, path is #{request.path}"
+    status = :ok
+    headers['Content-Type'] = 'text/plain'
+  elsif request.post? && request.path == '/create-post'
+    puts 'Got a new POST request!'
 
-    fields = URI.decode_www_form(message_body)
     post = Blog.new
-    fields.each do |name, value|
+    request.params.each do |name, value|
```
tomstuart, line 77: This is pretty much the only interesting piece of code we’re using! `#params` [merges](https://github.com/rack/rack/blob/2.2.3/lib/rack/request.rb#L469) the results of `#GET` and `#POST`, both of which [rely](https://github.com/rack/rack/blob/2.2.3/lib/rack/request.rb#L430) [upon](https://github.com/rack/rack/blob/2.2.3/lib/rack/request.rb#L454) `#parse_query`, which [delegates](https://github.com/rack/rack/blob/2.2.3/lib/rack/request.rb#L590) to [`Rack::QueryParser#parse_nested_query`](https://github.com/rack/rack/blob/2.2.3/lib/rack/query_parser.rb#L64), which is itself just a wrapper around [`#normalize_params`](https://github.com/rack/rack/blob/2.2.3/lib/rack/query_parser.rb#L83).

It isn’t super well documented, but by looking at the implementation of `#normalize_params` we can see that the `rack` gem actually supports a pretty sophisticated interpretation of name syntax:

```
>> parser = Rack::Utils.default_query_parser
=> #<Rack::QueryParser @params_class=Rack::QueryParser::Params, ...

>> parser.parse_nested_query('colours[]=red&colours[]=green&colours[]=blue')
=> {"colours"=>["red", "green", "blue"]}

>> parser.parse_nested_query('name[first]=Adrianna&name[last]=Chang')
=> {"name"=>{"first"=>"Adrianna", "last"=>"Chang"}}

>> parser.parse_nested_query('a[b][c][d][e]=42')
=> {"a"=>{"b"=>{"c"=>{"d"=>{"e"=>"42"}}}}}

>> parser.parse_nested_query('people[names][][first]=Tom&people[names][][last]=Stuart&people[names][][first]=Adrianna&people[names][][last]=Chang')
=> {"people"=>{"names"=>[{"first"=>"Tom", "last"=>"Stuart"}, {"first"=>"Adrianna", "last"=>"Chang"}]}}
```

None of this parameter-name parsing behaviour is part of any standard — HTML, and especially HTTP, don’t care what your form fields are called — but Rails builds upon it to support submission of form data that’s more structured than a flat list of key/value pairs. For example, this is what makes [the `collection_check_boxes` helper](https://guides.rubyonrails.org/form_helpers.html#the-collection-check-boxes-helper) work, and more generally these [parameter naming conventions](https://guides.rubyonrails.org/form_helpers.html#understanding-parameter-naming-conventions) construct a useful conspiracy between the form helpers and the controller which means that application developers don’t have to do as much work to reassemble structured data from a form submission.

tomstuart (approved):
This looks great! Not much to add since we paired on it, so just one minor digression into the guts of `rack` this time. 😁

challenge_9/rack_request.rb
adrianna-chang-shopify: This is super neat Tom - thanks for taking the time to share ❤️  
It's pretty wild that Rack is able to parse something like
```
>> parser.parse_nested_query('colours[]=red&colours[]=green&colours[]=blue')
=> {"colours"=>["red", "green", "blue"]}
```
! 

I know we talked previously about the fact that form keys don't have to be unique, so I was wondering how this would get parsed into a hash without the keys clobbering each other. Rack is pretty smart!

challenge_9/rack_request.rb
```diff
@@ -48,25 +52,29 @@
...
+    status = :ok
+    headers['Content-Type'] = 'text/html'
+  elsif request.get?
+    body << "method is #{request.request_method}, path is #{request.path}"
+    status = :ok
+    headers['Content-Type'] = 'text/plain'
+  elsif request.post? && request.path == '/create-post'
+    puts 'Got a new POST request!'
 
-    fields = URI.decode_www_form(message_body)
```
tomstuart: Oh hey, I missed that we don’t need to `require 'uri'` any more! 🤦🏻‍♂️

PR #9, opened 2021-03-24
Challenge 10 - Introducing `Rack::Response`

adrianna-chang-shopify:
In this challenge, we introduce a `Rack::Response` object to handle returning the Rack response, rather than building the array to be returned ourselves.

We also experiment with using the block version of `Response#finish`, which [takes a block](https://github.com/rack/rack/blob/master/lib/rack/response.rb#L80), and [returns self as the body](https://github.com/rack/rack/blob/master/lib/rack/response.rb#L89) of the Rack response. `Rack::Response` implements [#each](https://github.com/rack/rack/blob/master/lib/rack/response.rb#L98-L106), which takes a callback block and sets the response writer to be that callback. The app server can then pass a block to `each`, and this block gets called when we [`#write` to the response body](https://github.com/rack/rack/blob/master/lib/rack/response.rb#L115). This allows us to "stream" response content to the client (kind of like long polling) iteratively, rather than filling up a buffer and then having to terminate the response by calling `#finish`.

Example:

Let's say the app server has a `#send_response` method like this (this example is taken from https://github.com/macournoyer/tube/blob/master/tube.rb):
```ruby
    def send_response(env)
      status, headers, body = @app.call(env)
      reason = REASONS[status]

      @socket.write "HTTP/1.1 #{status} #{reason}\r\n"
      headers.each_pair do |name, value|
        @socket.write "#{name}: #{value}\r\n"
      end
      @socket.write "\r\n"
      body.each do |chunk|
        @socket.write chunk
      end
      body.close if body.respond_to? :close

      close
    end
```

And our Rack-compliant web app does this (example straight in this PR!):
```ruby
response.finish do
      response.write '<ul>'
      blog_data = store.transaction { store[:blogs] }
      loop do
        blog_data.each do |element|
          response.write '<li>'
          response.write "<strong>Title: #{CGI.escape_html(element.title)}</strong>, Content: #{CGI.escape_html(element.content)}"
          response.write '</li>'
        end
        sleep(1)
      end
     response.write '</ul>'
end
```

Then, in the `Rack::Response` object:
* `@block` is the block we send to `#finish` that writes to the response object in an infinite loop
* The `body` in the app server is actually our response instance
* Calling `body.each` in the app server sets the `@writer` to the block passed to `#each` (which is the block with `@socket.write chunk`)
* We then [call the `block`](https://github.com/rack/rack/blob/master/lib/rack/response.rb#L104), which performs `#write`s to the response object
* Now that the writer has actually been set to the app server's callback block, calling `response.write` actually invokes the `@socket.write chunk` proc
* This has us write straight to the socket! 😍

### DIFF
--- a/challenge_10/rack_response.rb
+++ b/challenge_10/rack_response.rb
@@ -1,0 +1,93 @@
+### Require dependencies
+require 'cgi'
+require 'rack/handler/puma'
+require 'rack'
+require 'uri'
+require 'yaml/store'
+
+### Status codes
+STATUS_CODES = {
+  ok: 200,
+  see_other: 303
+}
+
+# Struct to define what a Blog looks like
+Blog = Struct.new(:title, :content, keyword_init: true)
+
+def set_up_blogs_store(store_path)
+  store = YAML::Store.new(File.expand_path(store_path, __dir__))
+  store.transaction do
+    store[:blogs] = [] if store[:blogs].nil?
+  end
+  
+  # Seed some blog data
+  # Comment out if you'd like to start from scratch!
+  store.transaction do
+    if store[:blogs].empty?
+      store[:blogs] << Blog.new(title: 'My awesome blog!', content: 'my favourite HTML tags are <p> and <script>')
+      store[:blogs] << Blog.new(title: 'Another cool blog!', content: 'my favourite HTML tags are <br> and <hr>')
+    end
+  end
+  store
+end
+
+# Set up YAML store
+store = set_up_blogs_store('blogs.yml')
+
+app = lambda { |environment|
+  puts 'Rack app got a request!'
+  # Build Rack request
+  request = Rack::Request.new(environment)
+
+  headers = {}
+  body = []
+
+  if request.get? && request.path == '/show-data'
+    body << '<ul>'
+    blog_data = store.transaction { store[:blogs] }
+    blog_data.each do |element|
+      body << '<li>'
+      body << "<strong>Title: #{CGI.escape_html(element.title)}</strong>, Content: #{CGI.escape_html(element.content)}"
+      body << '</li>'
+    end
+    body << '</ul>'
+
+    # Prepare response
+    status = :ok
+    headers['Content-Type'] = 'text/html'
+  elsif request.get? && request.path == '/'
+    body << '<p><strong>Submit a new Blog Post!</p></strong>'
+    body << "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
+    body << "<p><label>Blog Title: <input name='title'></label></p>"
+    body << "<p><label>Content: <textarea name='content'></textarea></label></p>"
+    body << '<p><button>Submit post</button></p>'
+    body << '</form>'
+
+    # Prepare response
+    status = :ok
+    headers['Content-Type'] = 'text/html'
+  elsif request.get?
+    body << "method is #{request.request_method}, path is #{request.path}"
+    status = :ok
+    headers['Content-Type'] = 'text/plain'
+  elsif request.post? && request.path == '/create-post'
+    puts 'Got a new POST request!'
+
+    post = Blog.new
+    request.params.each do |name, value|
+      post[name] = value
+    end
+
+    store.transaction do
+      store[:blogs] << post
+    end
+
+    # Prepare response
+    status = :see_other
+    headers['Location'] = '/show-data' # NOTE: Don't need a Content-Type here, we're redirecting!
+  end
+
+  [STATUS_CODES[status], headers, body]
+}
+
+Rack::Handler::Puma.run(app, Port: 1234, Verbose: true)
--- a/challenge_10/rack_response.rb
+++ b/challenge_10/rack_response.rb
@@ -2,7 +2,6 @@
 require 'cgi'
 require 'rack/handler/puma'
 require 'rack'
-require 'uri'
 require 'yaml/store'
 
 # Struct to define what a Blog looks like
@@ -49,7 +48,7 @@ app = lambda { |environment|
         end
         sleep(1)
       end
-     response.write '</ul>'
+      response.write '</ul>'
     end
   elsif request.get? && request.path == '/'
     response.write '<p><strong>Submit a new Blog Post!</p></strong>'
--- a/challenge_10/rack_response.rb
+++ b/challenge_10/rack_response.rb
@@ -5,12 +5,6 @@ require 'rack'
 require 'uri'
 require 'yaml/store'
 
-### Status codes
-STATUS_CODES = {
-  ok: 200,
-  see_other: 303
-}
-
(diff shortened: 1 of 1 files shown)


challenge_10/rack_response.rb
```diff
@@ -83,11 +74,11 @@ def set_up_blogs_store(store_path)
     end
 
     # Prepare response
-    status = :see_other
-    headers['Location'] = '/show-data' # NOTE: Don't need a Content-Type here, we're redirecting!
+    response.redirect('/show-data', 303)
```
tomstuart, line 81: For the record: when we paired on this we discussed the pros & cons of replacing `:see_other` with the magic number `303`. On balance we felt it wasn’t perfect but also not a big deal since this is now the only place we need to set a non-default status code and so it felt appropriate to lose the `STATUS_CODES` hash. Maybe we’ll be able to revisit this issue in a future challenge. 🔮😄

challenge_10/rack_response.rb
```diff
@@ -39,37 +33,34 @@ def set_up_blogs_store(store_path)
   # Build Rack request
   request = Rack::Request.new(environment)
 
-  headers = {}
-  body = []
+  # Build Rack response
+  response = Rack::Response.new
```
tomstuart, line 36: There’s one more API style we didn’t try. We can [pass a block](https://github.com/rack/rack/blob/v2.2.2/lib/rack/response.rb#L65) to `Rack::Response.new` and do all the response generation inside that:

<pre>
>> Rack::Response.new { |response| response.write 'Hello, world!' }.finish
=> [200, {"Content-Length"=>"13"}, ["Hello, world!"]]
>> Rack::Response.new { <a href="https://medium.com/@baweaver/ruby-2-7-numbered-parameters-3f5c06a55fe4">_1</a>.redirect '/elsewhere' }.finish
=> [302, {"Location"=>"/elsewhere"}, []]
</pre>

I don’t think that would’ve been particularly useful in our case, but it is compact and convenient in some situations.

challenge_10/rack_response.rb
```diff
@@ -83,11 +74,11 @@ def set_up_blogs_store(store_path)
     end
 
     # Prepare response
-    status = :see_other
-    headers['Location'] = '/show-data' # NOTE: Don't need a Content-Type here, we're redirecting!
+    response.redirect('/show-data', 303)
   end
 
-  [STATUS_CODES[status], headers, body]
+  response.finish
```
tomstuart: One micro-convenience of `Rack::Request` is that it [aliases](https://github.com/rack/rack/blob/v2.2.2/lib/rack/response.rb#L96) `#finish` as `#to_a`, which makes splatting work:

```
>> response = Rack::Response.new
=> #<Rack::Response @status=200, @headers={}, @writer=#<Method: Rac...
>> response.write 'Hello, world!'
=> "Hello, world!"
>> status, headers, body = *response
=> [200, {"Content-Length"=>"13"}, ["Hello, world!"]]
>> status
=> 200
```

RACK TRIVIA: prior to [a change](https://github.com/rack/rack/commit/72959ebc2f300f3b2ccb7ae2aae9f199e611dfb6#diff-cfcc6d2645b6f77c5ec05acd448f280ce3004d3dba85b8aef3b3c81faf6d1373L79) in Rack v2.1.0, `Rack::Response` also aliased `#finish` as `#to_ary`, which made it possible to _implicitly_ coerce a response into an array, effectively making `#finish` optional everywhere:

```
>> response = Rack::Response.new
=> #<Rack::Response @status=200, @header={}, @writer=#<P...
>> response.write 'Hello, world!'
=> "Hello, world!"
>> status, headers, body = response
=> #<Rack::Response @status=200, @header={"Content-Lengt...
>> status
=> 200
```

But this no longer works. I guess Aaron felt it was a bit too implicit. 😄

challenge_10/rack_response.rb
```diff
@@ -58,9 +61,11 @@ def set_up_blogs_store(store_path)
 
     # Prepare response
     response.content_type = 'text/html'
+    response.finish
```
tomstuart, line 63: I suppose this is an example of a place where the combination of the `Rack::Response` block constructor and the pre-v2.1.0 behaviour was kind of convenient: if we didn’t care about explicitly calling `#finish` (i.e. to pass it a block so we could send a synchronous response), we could move all our response-generation logic inside `Rack::Response.new do |response| … end` and allow the response object to be returned from that, then let the application server implicitly call `#finish` (via `#to_ary`) when it tried to treat the response as an array. Instead we now have to be explicit about calling `#finish` and we don’t return the response at all, which in turn makes the block constructor form less elegant.

challenge_10/rack_response.rb
```diff
@@ -39,37 +33,34 @@ def set_up_blogs_store(store_path)
...
   request = Rack::Request.new(environment)
 
-  headers = {}
-  body = []
+  # Build Rack response
+  response = Rack::Response.new
 
   if request.get? && request.path == '/show-data'
-    body << '<ul>'
+   response.write '<ul>'
```
tomstuart: The indentation went a little funny again, but I can live with it 😂

challenge_10/rack_response.rb
```diff
@@ -0,0 +1,93 @@
+### Require dependencies
+require 'cgi'
+require 'rack/handler/puma'
+require 'rack'
+require 'uri'
```
tomstuart: My bad for [missing this](https://github.com/adrianna-chang-shopify/learning-project/pull/8/commits/db2c5efdd30369936e3e32e7e2d7dc5bdaa835f2#r601302364) in #8, but we’re no longer using `URI` because [`Rack::QueryParser`](https://github.com/rack/rack/blob/v2.2.2/lib/rack/query_parser.rb) does all the hard work for us.

tomstuart (approved):
Excellent! The code works, and the PR description is truly epic 🤩

challenge_10/rack_response.rb
adrianna-chang-shopify: Oooh foreshadowing 😆

challenge_10/rack_response.rb
adrianna-chang-shopify: Neat! Yeah I imagine this would be nice for responses that are quite compact, like maybe the `204 - No Content` status we looked at yesterday.

challenge_10/rack_response.rb
adrianna-chang-shopify: Bahahaha let me fix this

challenge_10/rack_response.rb
adrianna-chang-shopify: That makes a lot of sense! I noticed in [Aaron's PR](https://github.com/rack/rack/pull/1453) that the argument was "Responses are not arrays, so we should not allow them to be implicitly coerced to an array.", which makes sense too. It is convenient, but a little.. sneaky to have a response object get magically coerced into an array 😄

PR #10, opened 2021-03-31
Challenge 11 - Introduce a Database (SQLite 🎉)

adrianna-chang-shopify:
In this challenge, we introduce the `sqlite3` gem to use a SQLite database with our application, rather than writing to a YAML store to persist our blog data.

The [gem](https://github.com/sparklemotion/sqlite3-ruby) we use offers a Ruby interface for talking to the SQLite db engine, so we don't have to execute raw SQL in our application.

Something cool to note! The results of [`Database#execute`](https://github.com/sparklemotion/sqlite3-ruby/blob/d94c42b0b8e5005ee28cde75a58e15ea5670d1c4/lib/sqlite3/database.rb#L179) for a `SELECT` statement returns an array of arrays (row results).

```bash
irb(main):005:0> database = SQLite3::Database.new('application.sqlite3')
=> #<SQLite3::Database:0x00007ff989023048 @tracefunc=nil, @authorizer=nil, @encoding=nil, @busy...
irb(main):006:0> database.execute("SELECT * FROM blogs;")
=> [["My awesome blog!", "my favourite HTML tags are <p> and <script>"], ["Another cool blog!", "my favourite HTML tags are <br> and <hr>"]]
```

But, we can supply a [`results_as_hash` option](https://github.com/sparklemotion/sqlite3-ruby/blob/d94c42b0b8e5005ee28cde75a58e15ea5670d1c4/lib/sqlite3/database.rb#L99), and then we get an array of hashes, with the keys intelligently set to the column names:
```bash
irb(main):002:0> database = SQLite3::Database.new('application.sqlite3', results_as_hash: true)
=> #<SQLite3::Database:0x00007ff9b9908200 @tracefunc=nil, @authorizer=nil, @encoding=nil, @busy...
irb(main):003:0> database.execute("SELECT * FROM blogs;")
=> [{"title"=>"My awesome blog!", "content"=>"my favourite HTML tags are <p> and <script>"}, {"title"=>"Another cool blog!", "content"=>"my favourite HTML tags are <br> and <hr>"}]
```

This lets us do cool stuff with our `Blog` struct - ie. map over the results from the `execute` statement and implicitly coerce the hash into args for our Blog constructor:
```bash
irb(main):009:0> database = SQLite3::Database.new('application.sqlite3', results_as_hash: true)
=> #<SQLite3::Database:0x00007ff9cc82a3c0 @tracefunc=nil, @authorizer=nil, @encoding=nil, @busy...
irb(main):0010:0> Blog = Struct.new(:title, :content, keyword_init: true)
irb(main):0011:0> blogs = database.execute('SELECT * FROM blogs;').map { |data| Blog.new(data) }
=> [#<struct Blog title="My awesome blog!", content="my favourite HTML tags are <p> and <script...
```

Neat!


.gitignore
```diff
@@ -1 +1,2 @@
 blogs.yml
+application.sqlite3
```
adrianna-chang-shopify, line 2: I think we probably want to ignore this for the same reason we ignore our `blogs.yml` file 😁

tomstuart: I agree 😁

challenge_11/database.rb
```diff
@@ -0,0 +1,87 @@
...
+# Alternative
+# table_exists = database.get_first_value('SELECT COUNT(*) FROM sqlite_master WHERE type="table" AND name="blogs";')
+# if table_exists == 0
+#   database.execute('CREATE TABLE blogs (title VARCHAR(30), content TEXT);')
+# end
+
+# Seed some blog data if none exists already
+if database.get_first_value('SELECT COUNT(*) FROM blogs;') == 0
+  SEED_BLOGS.each do |blog|
+    database.execute('INSERT INTO blogs (title, content) VALUES (?, ?)', [blog.title, blog.content])
```
tomstuart, line 33: We could, of course, use the `blog.to_h` style here too.

challenge_11/database.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  # Build Rack response
+  response = Rack::Response.new
+
+  if request.get? && request.path == '/show-data'
+    response.content_type = 'text/html'
+    response.finish do
+      response.write '<ul>'
+      blogs = database.execute('SELECT * FROM blogs;').map { |data| Blog.new(data) }
+      blogs.each do |blog|
```
tomstuart, line 50: A small thing vaguely related to ideas we’ve discussed before: you can pass a block to `#execute`, in which case it [calls `#each` on the underlying `ResultSet`](https://github.com/sparklemotion/sqlite3-ruby/blob/v1.4.2/lib/sqlite3/database.rb#L198-L200) and yields each row to the block. This is interesting because `ResultSet#each` just [repeatedly calls `#next`](https://github.com/sparklemotion/sqlite3-ruby/blob/v1.4.2/lib/sqlite3/resultset.rb#L133) which itself [calls `#step`](https://github.com/sparklemotion/sqlite3-ruby/blob/v1.4.2/lib/sqlite3/resultset.rb#L108) on the statement object, which then [ultimately calls `sqlite3_step`](https://github.com/sparklemotion/sqlite3-ruby/blob/v1.4.2/ext/sqlite3/statement.c#L127) from the [SQLite C interface](https://www.sqlite.org/c3ref/step.html).

As you can see from the SQLite documentation, this is a sort of incremental API that is designed to be called repeatedly as the results are processed. I don’t know anything about the implementation details here, but it’s reasonable to assume that SQLite can break up the work of a large query and perform it in small chunks in response to repeated calls of `sqlite3_step`.

So my _point_ is that, when you call `#execute` with a block, the rows are kind of “streamed” back from the SQLite library as they’re generated, which would dovetail nicely with the [streaming responses](https://github.com/adrianna-chang-shopify/learning-project/commit/f5c3d6a2b5c93a3729e0effe8142f113db6cb28b) we’ve experimented with previously. It’s not hard to imagine connecting these features together to make a snappy-feeling web application that’s able to begin serving an expensive response immediately and stream the full body back to the client while it’s still being generated from the database. Neat.

challenge_11/database.rb
```diff
@@ -0,0 +1,87 @@
...
+    response.write "method is #{request.request_method}, path is #{request.path}"
+    response.content_type = 'text/plain'
+    response.finish
+  elsif request.post? && request.path == '/create-post'
+    puts 'Got a new POST request!'
+
+    blog = Blog.new(request.params)
+
+    # Write to database
+    database.execute('INSERT INTO blogs (title, content) VALUES (:title, :content)', blog.to_h)
```
tomstuart, line 78: If we wanted to go _really_ far down this rabbit hole, we could interpolate the result of `blog.members` into the SQL string to automatically populate the column names (and argument placeholder names, I guess) with the names of the fields from the `Struct`, and then it would be really DRY: we’d be fully reliant on the convention that the names of the `Blog` attributes perfectly match the column names of the `blogs` table. But that’s probably too much magic given that Active Record will imminently provide a far less hacky implementation of the same idea. 😁

tomstuart (approved):
Excellent work! 🎉

challenge_11/database.rb
adrianna-chang-shopify: Oh, this is really cool! Yeah, so I'd imagine that similar to what we were doing with the [streaming responses in challenge 10](https://github.com/adrianna-chang-shopify/learning-project/commit/f5c3d6a2b5c93a3729e0effe8142f113db6cb28b), we could call `#execute` with a block that would write the row to the response object, and if our response object was using `#finish` with the block form as well, this would result in that data getting written straight to the socket! I think it would look something like this?
```ruby
response.finish do
  response.write '<ul>'
  blogs = database.execute('SELECT * FROM blogs;') do |row|
    blog = Blog.new(row)
    response.write '<li>'
    response.write "<strong>Title: #{CGI.escape_html(blog.title)}</strong>, Content: #{CGI.escape_html(blog.content)}"
    response.write '</li>'
  end
  response.write '</ul>'
end
```

So if we had a million blogs in our db, this would be a really neat way to provide a responsive UX.

PR #11, opened 2021-04-05
Challenge 12 - Introducing the Active Record framework

adrianna-chang-shopify:
This challenge introduces the Active Record framework, which encapsulates database access and gives us a clean API for accessing our domain objects (ie. database rows) in a way that abstracts away the details of db queries / db access / etc.  

Note that Active Record hinges on the "Active record" design patten (who woulda thought?! 😂 ):
> an object that wraps a row in a database table or view, encapsulates the database access, and adds domain logic on that data

Although our `Blog` objects don't have any domain logic yet, this would be added directly to the Active Record class, offering us an API for both persistence logic and domain logic.

Note that
```ruby
ActiveRecord::Base.establish_connection(
  adapter:  "sqlite3",
  database: "application.sqlite3"
)

# Create the table if it doesn't exist, seed some data
ActiveRecord::Schema.define do
  unless table_exists?(:blogs)
    create_table :blogs do |t|
      t.string :title
      t.text :content
    end

    # Seed some data
    Blog.create!(title: 'My awesome blog!', content: 'my favourite HTML tags are <p> and <script>')
    Blog.create!(title: 'Another cool blog!', content: 'my favourite HTML tags are <br> and <hr>')
  end
end
```

is our bootstrapping code:
- `ActiveRecord::Base.establish_connection` is able to generate `application.sqlite3` if it doesn't exist
- [`ActiveRecord::Schema.define`](https://api.rubyonrails.org/classes/ActiveRecord/Schema.html#method-c-define) takes a block and makes all methods available to the current connection adapter available to the block. Consequently, we can use [`table_exists?`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L59-L63) and [`create_table` ](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L296-L341) from within the block to ensure we create the table in SQLite if it doesn't already exist
- Finally, we can use `#create!` to seed some blog data in our database

### DIFF
--- a/challenge_12/active_record.rb
+++ b/challenge_12/active_record.rb
@@ -1,0 +1,87 @@
+### Require dependencies
+require 'cgi'
+require 'rack/handler/puma'
+require 'rack'
+require 'sqlite3'
+
+# Struct to define what a Blog looks like
+Blog = Struct.new(:title, :content, keyword_init: true)
+
+SEED_BLOGS = [
+  Blog.new(title: 'My awesome blog!', content: 'my favourite HTML tags are <p> and <script>'),
+  Blog.new(title: 'Another cool blog!', content: 'my favourite HTML tags are <br> and <hr>')
+]
+
+# Create or open SQLite database
+database = SQLite3::Database.new('application.sqlite3', results_as_hash: true)
+
+# Check if we need to create our table
+existing_tables = database.execute('SELECT name FROM sqlite_master WHERE type="table";')
+unless existing_tables.include?({ 'name' => 'blogs'})
+  database.execute('CREATE TABLE blogs (title VARCHAR(30), content TEXT);')
+end
+
+# Alternative
+# table_exists = database.get_first_value('SELECT COUNT(*) FROM sqlite_master WHERE type="table" AND name="blogs";')
+# if table_exists == 0
+#   database.execute('CREATE TABLE blogs (title VARCHAR(30), content TEXT);')
+# end
+
+# Seed some blog data if none exists already
+if database.get_first_value('SELECT COUNT(*) FROM blogs;') == 0
+  SEED_BLOGS.each do |blog|
+    database.execute('INSERT INTO blogs (title, content) VALUES (?, ?)', [blog.title, blog.content])
+  end
+end
+
+app = lambda { |environment|
+  puts 'Rack app got a request!'
+  # Build Rack request
+  request = Rack::Request.new(environment)
+
+  # Build Rack response
+  response = Rack::Response.new
+
+  if request.get? && request.path == '/show-data'
+    response.content_type = 'text/html'
+    response.finish do
+      response.write '<ul>'
+      blogs = database.execute('SELECT * FROM blogs;').map { |data| Blog.new(data) }
+      blogs.each do |blog|
+        response.write '<li>'
+        response.write "<strong>Title: #{CGI.escape_html(blog.title)}</strong>, Content: #{CGI.escape_html(blog.content)}"
+        response.write '</li>'
+      end
+      response.write '</ul>'
+    end
+  elsif request.get? && request.path == '/'
+    response.write '<p><strong>Submit a new Blog Post!</p></strong>'
+    response.write "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
+    response.write "<p><label>Blog Title: <input name='title'></label></p>"
+    response.write "<p><label>Content: <textarea name='content'></textarea></label></p>"
+    response.write '<p><button>Submit post</button></p>'
+    response.write '</form>'
+
+    # Prepare response
+    response.content_type = 'text/html'
+    response.finish
+  elsif request.get?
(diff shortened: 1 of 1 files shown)


challenge_12/active_record.rb
```diff
@@ -46,8 +28,8 @@
     response.content_type = 'text/html'
     response.finish do
       response.write '<ul>'
-      blogs = database.execute('SELECT * FROM blogs;').map { |data| Blog.new(data) }
-      blogs.each do |blog|
+
+      Blog.all.each do |blog|
```
tomstuart, line 47: Another entry in my apparently endless series of inconsequential remarks about streaming: we could use [`.find_each`](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each) here instead if we wanted to incrementally render each post into the response rather than fetching all of them into an array and then iterating. Under the hood this retrieves rows from the database [in batches](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_in_batches) (default 1000 per batch) so that we’re never trying to fit the entire table’s data in memory. There’s a trade-off here that might require some tuning in production: larger batches take longer to retrieve and use more memory, but there may be some amount of per-query overhead (e.g. network round-trip time to the database server) which we can reduce by using fewer, larger batches.

challenge_12/active_record.rb
```diff
@@ -72,10 +54,8 @@
   elsif request.post? && request.path == '/create-post'
     puts 'Got a new POST request!'
 
-    blog = Blog.new(request.params)
-
-    # Write to database
-    database.execute('INSERT INTO blogs (title, content) VALUES (:title, :content)', blog.to_h)
+    # The request params can be passed directly to the Blog's create method - neat!
+    blog = Blog.create!(request.params)
```
tomstuart, line 73: Another trade-off here: we’re now assuming that _every_ key in `request.params` is a valid attribute name for the `Blog` model, and `.create!` will raise an `ActiveModel::UnknownAttributeError` exception if it isn’t. This is convenient in some ways because we’re establishing the convention that every parameter to this `POST` request has a matching attribute name on the `Blog` model, which means we don’t need to change this part of the code at all if we decide to add a new column to the table and a corresponding field to the form; but, on the other hand, it’s less explicit about what’s going on, and less tolerant in cases where e.g. an extra parameter gets passed by mistake or for some other (perhaps nefarious) purpose.

Rails provides the [strong parameters](https://guides.rubyonrails.org/action_controller_overview.html#strong-parameters) API in an attempt to give us the best of both worlds: we can still write convenient code like `Blog.create!(params)`, but we must explicitly say which parameters are allowed through.

challenge_12/active_record.rb
```diff
@@ -11,11 +11,26 @@ class Blog < ActiveRecord::Base
...
+    create_table :blogs do |t|
+      t.string :title
+      t.text :content
+    end
+
+    # Seed some data
+    Blog.create!(title: 'My awesome blog!', content: 'my favourite HTML tags are <p> and <script>')
+    Blog.create!(title: 'Another cool blog!', content: 'my favourite HTML tags are <br> and <hr>')
+  end
+end
```
tomstuart, line 32: This is super cool! TIL, thank you! 🎉

tomstuart (approved):
Excellent. And I’m happy to have learned how to use `ActiveRecord::Scheme.define`! 😁

challenge_12/active_record.rb
adrianna-chang-shopify: Your "inconsequential" remarks are my favourite part! 😆  ❤️ 
Oooh, yes I've encountered `.find_each` a couple of times now. I always forget that it's a nice option when the Active Record collection might be very large.

challenge_12/active_record.rb
adrianna-chang-shopify: Ah, great point! Let me incorporate the use of strong params here ❤️

tomstuart: That might be jumping the gun a little because we’ve not introduced Action Controller yet. So feel free to take a look, but off the top of my head I don’t know how easy it’ll be!

tomstuart: (But Action Controller is coming up real soon now…)

adrianna-chang-shopify: Right!! I was going to experiment with using strong parameters in this PR, but that would require transforming the request params into `ActionController::Parameters` and I suppose that would be jumping ahead a bit huh? 😂

tomstuart: A little, and you’ll get that for free in the next challenge or two. So no harm trying, but don’t get bogged down in it is all.

adrianna-chang-shopify: Hahaha I actually wrote my comment before seeing your next two comments (I guess we were reading each others minds 😉 ) - I'll hold off until the Action Controller challenges I think!

PR #12, opened 2021-04-21
Challenge 13 - Use `Action Dispatch` to route requests

adrianna-chang-shopify:
In this challenge, we introduce the Action Dispatch framework to handle parsing our incoming HTTP requests and routing them to the correct endpoint.

But what is an endpoint?!
- Conforms to the same specification as a Rack application, in that it expects to be called with an `environment` hash and then return a three-element array
- We initialize an [`ActionDispatch::Routing::RouteSet`](https://www.rubydoc.info/docs/rails/ActionDispatch/Routing/RouteSet), which can draw routes using a `#draw` method that takes a block declaring the routes to draw
- The[ DSL for mapping routes ](https://www.rubydoc.info/docs/rails/ActionDispatch/Routing/Mapper/HttpHelpers) uses `<action> <path>, to: <endpoint>`, where `action` is `get`, `put`, `post`, etc...

The `RouteSet` is super duper smart and can even route based on wildcard patterns:
```ruby
      match '*path', via: :all, to: -> environment {
        response = Rack::Response.new
        response.write "Sorry, I don’t know what #{environment['PATH_INFO']} is"
        response.content_type = 'text/plain'
        response.status = 404
        response.finish
      }
```

**BIG TAKEAWAY!**
> The end result is to stitch several small Rack applications (“endpoints”) together into a single large Rack application which can serve responses by directing requests to the correct endpoint.

The router essentially stitches together these mini Rack apps (endpoints), and then provides a `.call` method that takes the `environment`, and defers the work to the appropriate endpoint.

We can wrap the call to the router inside _another_ Rack application like so: 
```ruby
app = lambda { |environment| router.call(environment) }
```
This allows us to decorate the app with additional behaviour by wrapping it in middleware-esque classes! See https://github.com/adrianna-chang-shopify/learning-project/commit/f2dbf9921a1f214b50d4bc0dc0d145ee475bc79e for all the fun Tom and I had implementing these 🎉 

**config.ru fun 🦘**
https://github.com/adrianna-chang-shopify/learning-project/commit/34e2322dbec017fac6d9ef8245aa931f21c21c11 refactors our main app file to actually have an invokable endpoint that we can call from config.ru. @tomstuart let me know if this is an okay way to do this! To sum, we:
- Make our rack application an actual Ruby object in the `action_dispatch.rb` file - on initialize, it creates a routeset and draws the routes we want
- It implements `call`, which takes an environment argument and returns the results of calling our router (conforms to the Rack specification)
- In `config.ru`, we require our app file, implement our middleware logic, and then use `run MyApp.new`

Now, we can just use `puma` from the console and it all works! 🎉  (I couldn't figure out how to specify the port number / verbose mode when using a config.ru 🤔 )

### DIFF
--- a/challenge_13/action_dispatch.rb
+++ b/challenge_13/action_dispatch.rb
@@ -1,4 +1,5 @@
 ### Require dependencies
+require 'action_dispatch'
 require 'active_record'
 require 'cgi'
 require 'rack/handler/puma'
@@ -23,15 +24,21 @@ ActiveRecord::Schema.define do
   end
 end
 
-app = lambda { |environment|
-  puts 'Rack app got a request!'
-  # Build Rack request
-  request = Rack::Request.new(environment)
-
-  # Build Rack response
-  response = Rack::Response.new
-
-  if request.get? && request.path == '/show-data'
+router = ActionDispatch::Routing::RouteSet.new
+router.draw do
+  get '/', to: -> environment {
+    response = Rack::Response.new
+    response.write '<p><strong>Submit a new Blog Post!</p></strong>'
+    response.write "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
+    response.write "<p><label>Blog Title: <input name='title'></label></p>"
+    response.write "<p><label>Content: <textarea name='content'></textarea></label></p>"
+    response.write '<p><button>Submit post</button></p>'
+    response.write '</form>'
+    response.content_type = 'text/html'
+    response.finish
+  }
+  get '/show-data', to: -> environment {
+    response = Rack::Response.new
     response.content_type = 'text/html'
     response.finish do
       response.write '<ul>'
@@ -43,31 +50,28 @@ app = lambda { |environment|
       end
       response.write '</ul>'
     end
-  elsif request.get? && request.path == '/'
-    response.write '<p><strong>Submit a new Blog Post!</p></strong>'
-    response.write "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
-    response.write "<p><label>Blog Title: <input name='title'></label></p>"
-    response.write "<p><label>Content: <textarea name='content'></textarea></label></p>"
-    response.write '<p><button>Submit post</button></p>'
-    response.write '</form>'
-
-    # Prepare response
-    response.content_type = 'text/html'
-    response.finish
-  elsif request.get?
-    response.write "method is #{request.request_method}, path is #{request.path}"
-    response.content_type = 'text/plain'
-    response.finish
-  elsif request.post? && request.path == '/create-post'
+  }
+  post 'create-post', to: -> environment {
+    response = Rack::Response.new
+    request = Rack::Request.new(environment)
     puts 'Got a new POST request!'
 
-    # The request params can be passed directly to the Blog's create method - neat!
-    blog = Blog.create!(request.params)
-
-    # Prepare response
+    Blog.create!(request.params)
     response.redirect('/show-data', 303)
     response.finish
-  end
+  }
+  match '*path', via: :all, to: -> environment {
+    response = Rack::Response.new
+    response.write "Sorry, I don’t know what #{environment['PATH_INFO']} is"
+    response.content_type = 'text/plain'
+    response.status = 404
+    response.finish
+  }
+end
+
+app = lambda { |environment|
+  puts 'Rack app got a request!'
+  router.call(environment) 
 }
 
 # Run the Puma app server with our web application
--- a/challenge_13/action_dispatch.rb
+++ b/challenge_13/action_dispatch.rb
@@ -69,10 +69,34 @@ router.draw do
   }
 end
 
-app = lambda { |environment|
-  puts 'Rack app got a request!'
-  router.call(environment) 
-}
+class LoggerMiddleware
+  def initialize(app, string)
+    @app = app
+    @string = string
+  end
+
+  def call(environment)
+    puts @string
+    @app.call(environment)
+  end
+end
+
+class CharsetMiddleware
+  def initialize(app, charset: 'UTF-8')
+    @app = app
+    @charset = charset
+  end
+
+  def call(environment)
+    response = Rack::Response[*@app.call(environment)]
+    response.content_type = "#{response.content_type}; charset=#{@charset}"
+    response.finish
+  end
+end
+
+app = lambda { |environment| router.call(environment) }
+app = CharsetMiddleware.new(app)
+app = LoggerMiddleware.new(app, 'Rack app got a request!')
 
 # Run the Puma app server with our web application
 Rack::Handler::Puma.run(app, Port: 1234, Verbose: true)
--- a/challenge_13/action_dispatch.rb
+++ b/challenge_13/action_dispatch.rb
@@ -2,7 +2,6 @@
 require 'action_dispatch'
 require 'active_record'
 require 'cgi'
-require 'rack/handler/puma'
 require 'rack'
 
 class Blog < ActiveRecord::Base; end
@@ -74,7 +73,8 @@ class MyApp
       }
       match '*path', via: :all, to: -> environment {
         response = Rack::Response.new
-        response.write "Sorry, I don’t know what #{environment['PATH_INFO']} is"
+        request = Rack::Request.new(environment)
+        response.write "Sorry, I don’t know what #{request.path_info} is"
         response.content_type = 'text/plain'
         response.status = 404
         response.finish
--- a/challenge_13/action_dispatch.rb
+++ b/challenge_13/action_dispatch.rb
@@ -24,79 +24,61 @@ ActiveRecord::Schema.define do
   end
 end
 
-router = ActionDispatch::Routing::RouteSet.new
-router.draw do
-  get '/', to: -> environment {
-    response = Rack::Response.new
-    response.write '<p><strong>Submit a new Blog Post!</p></strong>'
-    response.write "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
-    response.write "<p><label>Blog Title: <input name='title'></label></p>"
-    response.write "<p><label>Content: <textarea name='content'></textarea></label></p>"
-    response.write '<p><button>Submit post</button></p>'
-    response.write '</form>'
-    response.content_type = 'text/html'
-    response.finish
-  }
-  get '/show-data', to: -> environment {
-    response = Rack::Response.new
-    response.content_type = 'text/html'
-    response.finish do
-      response.write '<ul>'
+class MyApp
+  def initialize
+    @router = ActionDispatch::Routing::RouteSet.new
+    draw_routes
+  end
 
-      Blog.all.each do |blog|
-        response.write '<li>'
-        response.write "<strong>Title: #{CGI.escape_html(blog.title)}</strong>, Content: #{CGI.escape_html(blog.content)}"
-        response.write '</li>'
-      end
-      response.write '</ul>'
+  def call(environment)
+    @router.call(environment)
+  end
+
+  private
+
+  def draw_routes
+    @router.draw do
+      get '/', to: -> environment {
+        response = Rack::Response.new
+        response.write '<p><strong>Submit a new Blog Post!</p></strong>'
+        response.write "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
+        response.write "<p><label>Blog Title: <input name='title'></label></p>"
+        response.write "<p><label>Content: <textarea name='content'></textarea></label></p>"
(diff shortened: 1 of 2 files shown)


challenge_13/action_dispatch.rb
```diff
@@ -43,31 +50,28 @@ class Blog < ActiveRecord::Base; end
...
-    response.content_type = 'text/html'
+    Blog.create!(request.params)
+    response.redirect('/show-data', 303)
     response.finish
-  elsif request.get?
-    response.write "method is #{request.request_method}, path is #{request.path}"
+  }
+  match '*path', via: :all, to: -> environment {
+    response = Rack::Response.new
+    response.write "Sorry, I don’t know what #{environment['PATH_INFO']} is"
```
tomstuart: I missed that this code (copied from my doc, I think) is referencing a key in the `environment` hash directly. It’d be nicer to access the hash indirectly through `Rack::Request` for the same “less knowledge is better” reason we started doing that earlier.

challenge_13/action_dispatch.rb
```diff
@@ -23,15 +24,21 @@ class Blog < ActiveRecord::Base; end
...
 
-app = lambda { |environment|
-  puts 'Rack app got a request!'
-  # Build Rack request
-  request = Rack::Request.new(environment)
-
-  # Build Rack response
-  response = Rack::Response.new
-
-  if request.get? && request.path == '/show-data'
```
tomstuart: Ultra nitpick: it was a little hard to follow this diff because the endpoints changed order, e.g. previously `/show-data` was first but now `/` is first. If rearranging these was important to you then you could’ve first made a commit reordering the code as-is and then a separate commit converting to `ActionDispatch::Routing::RouteSet` (or vice versa) rather than combining both changes. No big deal though, I got there in the end 😄

challenge_13/action_dispatch.rb
```diff
@@ -69,10 +69,34 @@ class Blog < ActiveRecord::Base; end
...
+  end
+
+  def call(environment)
+    response = Rack::Response[*@app.call(environment)]
+    response.content_type = "#{response.content_type}; charset=#{@charset}"
+    response.finish
+  end
+end
+
+app = lambda { |environment| router.call(environment) }
```
tomstuart: This line gets removed in the next commit anyway, but note that it’s effectively the same as `app = router` because `lambda { |*args| whatever.call(*args) }` adds nothing and is therefore indistinguishable from just `whatever` (mathematically they are [extensionally equal](https://en.wikipedia.org/wiki/Extensionality) 🤓 ). So in practice you could’ve removed this line entirely and just done `app = CharsetMiddleware.new(router)` below, or indeed written `app = router` if you wanted to make it explicit that the router itself _is_ the application at the bottom of the middleware stack.

challenge_13/config.ru
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def call(environment)
+    response = Rack::Response[*@app.call(environment)]
+    response.content_type = "#{response.content_type}; charset=#{@charset}"
+    response.finish
+  end
+end
+
+use LoggerMiddleware, 'Rack app got a request!'
```
tomstuart, line 28: lol, guess that “extra constructor parameters” thing works 😅

challenge_13/action_dispatch.rb
```diff
@@ -69,10 +69,34 @@ class Blog < ActiveRecord::Base; end
...
+
+class CharsetMiddleware
+  def initialize(app, charset: 'UTF-8')
+    @app = app
+    @charset = charset
+  end
+
+  def call(environment)
+    response = Rack::Response[*@app.call(environment)]
+    response.content_type = "#{response.content_type}; charset=#{@charset}"
```
tomstuart: As a thought experiment: how could we make this middleware [liberal in what it accepts](https://en.wikipedia.org/wiki/Robustness_principle) by having it deal gracefully with situations where the response’s `Content-Type` header already contains a `charset` [media type parameter](https://tools.ietf.org/html/rfc7231#section-3.1.1.1)? ([Hint](https://github.com/rack/rack/blob/v2.2.2/lib/rack/response.rb#L195-L197).)

challenge_13/action_dispatch.rb
```diff
@@ -69,10 +69,34 @@ class Blog < ActiveRecord::Base; end
...
+  def initialize(app, string)
+    @app = app
+    @string = string
+  end
+
+  def call(environment)
+    puts @string
+    @app.call(environment)
+  end
+end
```
tomstuart: If you’re curious (😂 ), check out [`Rack::CommonLogger`](https://github.com/rack/rack/blob/v2.2.2/lib/rack/common_logger.rb) which implements a more sophisticated version of the same idea. In particular I think you’ll be interested in their technique for solving the problem of “how do we write a log message when the request is completely over, given that (because of streaming responses) that might be much later than the time when `#call` actually returns?”. Note also the mechanism by which it integrates with the confusingly-named [`Rack::Logger`](https://github.com/rack/rack/blob/v2.2.2/lib/rack/logger.rb), whose job is to _provide_ a logger to subsequent middleware.

challenge_13/action_dispatch.rb
```diff
@@ -69,10 +69,34 @@ class Blog < ActiveRecord::Base; end
...
+    @app = app
+    @charset = charset
+  end
+
+  def call(environment)
+    response = Rack::Response[*@app.call(environment)]
+    response.content_type = "#{response.content_type}; charset=#{@charset}"
+    response.finish
+  end
+end
```
tomstuart: Again, you have independently invented something pretty similar to [`Rack::Charset`](https://github.com/adamluzsi/rack-charset/blob/v1.0.0/lib/rack/charset.rb) here. Conventions work! 😄

challenge_13/action_dispatch.rb
```diff
@@ -24,79 +24,61 @@ class Blog < ActiveRecord::Base; end
...
+class MyApp
+  def initialize
+    @router = ActionDispatch::Routing::RouteSet.new
+    draw_routes
   end
 
   def call(environment)
-    puts @string
-    @app.call(environment)
+    @router.call(environment)
```
tomstuart, line 33: What you’ve done works well, but note that there is a bit of optional boilerplate here — `MyApp#call` doesn’t do anything useful, it’s just delegating to `@router.call`. So you could just as well have said `MyApp = ActionDispatch::Routing::RouteSet.new; MyApp.draw do … end` and cut out the need to instantiate a class altogether. (Of course, implementing the app as a class gives you the advantages of being able to parameterise its constructor and make multiple instances of it in a single process, neither of which you’re doing right now but could be desirable in other situations.)

challenge_13/action_dispatch.rb
```diff
@@ -24,79 +24,61 @@ class Blog < ActiveRecord::Base; end
...
+    end
   end
 end
-
-app = lambda { |environment| router.call(environment) }
-app = CharsetMiddleware.new(app)
-app = LoggerMiddleware.new(app, 'Rack app got a request!')
-
-# Run the Puma app server with our web application
-Rack::Handler::Puma.run(app, Port: 1234, Verbose: true)
```
tomstuart: As you noted, we’ve lost the ability to control the port number and verbosity ourselves now that we’re no longer starting Puma, but that’s kind of a good thing because instead it’s up to the person _running the application_:

```
% puma --help
puma <options> <rackup file>
    -b, --bind URI                   URI to bind to (tcp://, unix://, ssl://)
        --bind-to-activated-sockets [only]
                                     Bind to all activated sockets
    -C, --config PATH                Load PATH as a config file
        --control-url URL            The bind url to use for the control server. Use 'auto' to use temp unix server
        --control-token TOKEN        The token to use as authentication for the control server
        --debug                      Log lowlevel debugging information
        --dir DIR                    Change to DIR before starting
    -e, --environment ENVIRONMENT    The environment to run the Rack app on (default development)
    -f, --fork-worker=[REQUESTS]     Fork new workers from existing worker. Cluster mode only
                                     Auto-refork after REQUESTS (default 1000)
    -I, --include PATH               Specify $LOAD_PATH directories
    -p, --port PORT                  Define the TCP port to bind to
                                     Use -b for more advanced options
        --pidfile PATH               Use PATH as a pidfile
        --preload                    Preload the app. Cluster mode only
        --prune-bundler              Prune out the bundler env if possible
        --extra-runtime-dependencies GEM1,GEM2
                                     Defines any extra needed gems when using --prune-bundler
    -q, --quiet                      Do not log requests internally (default true)
    -v, --log-requests               Log requests as they occur
    -R, --restart-cmd CMD            The puma command to run during a hot restart
                                     Default: inferred
    -S, --state PATH                 Where to store the state details
    -t, --threads INT                min:max threads to use (default 0:16)
        --early-hints                Enable early hints support
    -V, --version                    Print the version information
    -w, --workers COUNT              Activate cluster mode: How many worker processes to create
        --tag NAME                   Additional text to display in process listing
        --redirect-stdout FILE       Redirect STDOUT to a specific file
        --redirect-stderr FILE       Redirect STDERR to a specific file
        --[no-]redirect-append       Append to redirected files
    -h, --help                       Show help
```

So a user who doesn’t care can type simply `puma` and get the defaults, but another can say `puma --port 1234 --log-requests` if they want the previous behaviour.

challenge_13/config.ru
```diff
@@ -0,0 +1,31 @@
...
+    @app = app
+    @charset = charset
+  end
+
+  def call(environment)
+    response = Rack::Response[*@app.call(environment)]
+    response.content_type = "#{response.content_type}; charset=#{@charset}"
+    response.finish
+  end
+end
```
tomstuart: To state the obvious: you could easily move these pieces of middleware into their own file(s) and `require_relative` them if you wanted to keep `config.ru` small. But since this “configuration” file is literally just Ruby, it’s also fine to have as much code here as you like!

challenge_13/action_dispatch.rb
tomstuart: psst, now we can remove another `require` 🔥

tomstuart (approved):
Another excellent and thorough solution, well done! I particularly love how the combination of `ActiveRecord::Schema.define` and `config.ru` now makes this app incredibly easy to run: just type `puma` and you’re cooking. 🎉

challenge_13/action_dispatch.rb
adrianna-chang-shopify: Yes, my bad! I chose to refactor `/` first because it seemed like the simplest scenario to migrate over, and forgot that would mess up the diffs 😅

challenge_13/action_dispatch.rb
adrianna-chang-shopify: Verrrry coool 🤓  Lambdas are pretty mind-bendy! Being able to just assign `app = router` as the entrypoint to the middleware stack is pretty neat 🎉

challenge_13/action_dispatch.rb
adrianna-chang-shopify: Oh yay!!! 🙌

challenge_13/action_dispatch.rb
adrianna-chang-shopify: Ooh, the use of a `BodyProxy` is really interesting! Definitely on theme with our explorations into streaming responses - it's neat that the Common Logger middleware returns a body that is actual a callable BodyProxy, configured with a block, that can be invoked whenever the response is sent back to the client, at a much later point in time to our initial call to the Common Logger middleware!

challenge_13/action_dispatch.rb
adrianna-chang-shopify: Right, this is handy to know - thanks!

challenge_13/action_dispatch.rb
adrianna-chang-shopify: Let me give it a shot! 😉

challenge_13/config.ru
```diff
@@ -20,7 +20,7 @@ class CharsetMiddleware
 
   def call(environment)
     response = Rack::Response[*@app.call(environment)]
-    response.content_type = "#{response.content_type}; charset=#{@charset}"
+    response.content_type = "#{response.content_type}; charset=#{@charset}" unless response.media_type_params.any?
```
tomstuart, line 23: Great! Note that, in general, other media type parameters (e.g. [`boundary`](https://tools.ietf.org/html/rfc2046#section-5.1.1) in the case of `multipart/mixed`) might be present, so the _super_ correct thing here is to check for the presence of the `charset` parameter [specifically](https://github.com/adamluzsi/rack-charset/blob/v1.0.0/lib/rack/charset.rb#L16), but in practice other parameters are very rare so what you’ve done here will work fine.

adrianna-chang-shopify: Ah, yep that makes sense! Thanks for making note of this ❤️

PR #13, opened 2021-04-29
Challenge 14 - Introducing Action Controller!

adrianna-chang-shopify:
In this challenge, we introduce the Action Controller library. It offers us a way to organize our Rack endpoints in a class (as methods - but note that these are kind of "overloaded" method definitions), and gives us some nice helper methods (`#params`, `#render`, etc...) so that we no longer have to do the work of creating Rack request and response objects ourselves.

Note that, unlike typical OO programming, these controller methods are _not_ intended to be used on an instance of a controller class, but are always intended to be used by calling `MyController.action(:<method_name>)`. It is this `.action` method that actually transforms the contents of the method into a Rack endpoint (ie. a Proc capable of receiving an environment hash, and returning the three-element array that Rack endpoints must return).

### Perks of using Action Controller endpoints
- Don't have to create Rack request / response objects anymore!
- Can combine the content type and status into a single line with a more human-readable DSL, ie: `render plain: response, status: :not_found`
- Similar to the above, we now have helpers for things like status codes (`status: :see_other`, etc...), so we can use those without needing to remember the numbers corresponding to different status codes!

### Key Takeaways from Tom
- The choice to have controller endpoints constructed as overloaded method definitions was simply a design design made by humans! This could easily have been some other DSL (ie. how [Sinatra defines endpoints](http://sinatrarb.com/))
- Always need to call `html_safe` on the response string if we are storing raw HTML in it, because Action Controller's default is to escape any HTML contained in the response body
- The Action Controller endpoints return the body as an `ActionDispatch::Response`, and these include `Rack::Request::Helpers`, which allows us to still use `request.path_info`:
```ruby
  def not_found
    response = "Sorry, I don’t know what #{request.path_info} is"
    render plain: response, status: :not_found
  end
```
- We got to experiment with the [Strong Parameters API](https://api.rubyonrails.org/classes/ActionController/StrongParameters.html) here:
```ruby
  def create_post

    Blog.create!(params.permit(:title, :content))
    redirect_to "/show-data", status: :see_other
  end
```
If we just passed the parameters directly into `Blog.create!`, we would raise an `ActiveModel::ForbiddenAttributesError ` exception because it'd be using mass assignment without explicitly permitting those params.

### DIFF
--- a/challenge_14/action_controller.rb
+++ b/challenge_14/action_controller.rb
@@ -1,0 +1,84 @@
+### Require dependencies
+require 'action_dispatch'
+require 'active_record'
+require 'cgi'
+require 'rack'
+
+class Blog < ActiveRecord::Base; end
+
+ActiveRecord::Base.establish_connection(
+  adapter:  "sqlite3",
+  database: "application.sqlite3"
+)
+
+# Create the table if it doesn't exist, seed some data
+ActiveRecord::Schema.define do
+  unless table_exists?(:blogs)
+    create_table :blogs do |t|
+      t.string :title
+      t.text :content
+    end
+    Blog.create!(title: 'My awesome blog!', content: 'my favourite HTML tags are <p> and <script>')
+    Blog.create!(title: 'Another cool blog!', content: 'my favourite HTML tags are <br> and <hr>')
+  end
+end
+
+class MyApp
+  def initialize
+    @router = ActionDispatch::Routing::RouteSet.new
+    draw_routes
+  end
+
+  def call(environment)
+    @router.call(environment)
+  end
+
+  private
+
+  def draw_routes
+    @router.draw do
+      get '/', to: -> environment {
+        response = Rack::Response.new
+        response.write '<p><strong>Submit a new Blog Post!</p></strong>'
+        response.write "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
+        response.write "<p><label>Blog Title: <input name='title'></label></p>"
+        response.write "<p><label>Content: <textarea name='content'></textarea></label></p>"
+        response.write '<p><button>Submit post</button></p>'
+        response.write '</form>'
+        response.content_type = 'text/html'
+        response.finish
+      }
+      get '/show-data', to: -> environment {
+        response = Rack::Response.new
+        response.content_type = 'text/html'
+        response.finish do
+          response.write '<ul>'
+    
+          Blog.all.each do |blog|
+            response.write '<li>'
+            response.write "<strong>Title: #{CGI.escape_html(blog.title)}</strong>, Content: #{CGI.escape_html(blog.content)}"
+            response.write '</li>'
+          end
+          response.write '</ul>'
+        end
+      }
+      post 'create-post', to: -> environment {
+        response = Rack::Response.new
+        request = Rack::Request.new(environment)
+        puts 'Got a new POST request!'
+    
+        Blog.create!(request.params)
+        response.redirect('/show-data', 303)
+        response.finish
+      }
+      match '*path', via: :all, to: -> environment {
+        response = Rack::Response.new
+        request = Rack::Request.new(environment)
+        response.write "Sorry, I don’t know what #{request.path_info} is"
+        response.content_type = 'text/plain'
+        response.status = 404
+        response.finish
+      }
+    end
+  end
+end
--- a/challenge_14/action_controller.rb
+++ b/challenge_14/action_controller.rb
@@ -1,4 +1,5 @@
 ### Require dependencies
+require 'action_controller'
 require 'action_dispatch'
 require 'active_record'
 require 'cgi'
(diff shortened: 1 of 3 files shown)


challenge_14/action_controller.rb
```diff
@@ -37,48 +76,10 @@ def call(environment)
 
   def draw_routes
     @router.draw do
-      get '/', to: -> environment {
-        response = Rack::Response.new
```
tomstuart: Action Controller is taking care of all the Rack request/response business now, so we needn’t `require 'rack'` any more.

challenge_14/action_controller.rb
```diff
@@ -23,6 +24,44 @@ class Blog < ActiveRecord::Base; end
...
+  def create_post
+    puts 'Got a new POST request!'
+
+    Blog.create!(params.permit(:title, :content))
+    redirect_to "/show-data", status: :see_other
+  end
+
+  def not_found
+    response = "Sorry, I don’t know what #{request.path_info} is"
+    render plain: response, status: :not_found
```
tomstuart: Are we still reliant upon `CharsetMiddleware` for this [`ActionDispatch::Response`](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/http/response.rb) to be rendered correctly?

challenge_14/action_controller.rb
```diff
@@ -23,6 +24,44 @@ class Blog < ActiveRecord::Base; end
   end
 end
 
+class AppController < ActionController::Base
+  def root
+    response = ""
```
tomstuart: The only thing to be slightly cautious of here is that `#response` is [an instance method](https://api.rubyonrails.org/classes/ActionController/Base.html#method-i-response) on `ActionController::Base`, so using the same name for a local variable is preventing you from calling it (with implicit receiver) if you want to. Not an actual problem, but another choice like `response_body` might be safer.

challenge_14/action_controller.rb
```diff
@@ -23,6 +24,44 @@ class Blog < ActiveRecord::Base; end
...
+      response += "</li>"
+    end
+    response += "</ul>"
+    render html: response.html_safe
+  end
+
+  def create_post
+    puts 'Got a new POST request!'
+
+    Blog.create!(params.permit(:title, :content))
```
tomstuart, line 56: As we discussed when pairing, this could also be `Blog.create!(title: params[:title], content: params[:content])`, because that effectively achieves the same goal of preventing any unintended parameters from slipping through. `#permit` is tidier though!

challenge_14/action_controller.rb
```diff
@@ -37,48 +76,10 @@ def call(environment)
...
-      }
-      match '*path', via: :all, to: -> environment {
-        response = Rack::Response.new
-        request = Rack::Request.new(environment)
-        response.write "Sorry, I don’t know what #{request.path_info} is"
-        response.content_type = 'text/plain'
-        response.status = 404
-        response.finish
-      }
+      get '/', to: AppController.action(:root)
```
tomstuart: I should’ve mentioned this [last time](https://github.com/adrianna-chang-shopify/learning-project/pull/12), but `ActionDispatch::Routing::Mapper` has a [`#root` method](https://github.com/rails/rails/blob/v6.1.3.1/actionpack/lib/action_dispatch/routing/mapper.rb#L1680-L1698) which is [shorthand for](https://github.com/rails/rails/blob/v6.1.3.1/actionpack/lib/action_dispatch/routing/mapper.rb#L1989-L1992) `match '/', via: :get`, so you could use that here if you’d like.

challenge_14/action_controller.rb
```diff
@@ -23,6 +24,44 @@ class Blog < ActiveRecord::Base; end
...
+    render html: response.html_safe
+  end
+
+  def show_data
+    response = ""
+    response += "<ul>"
+
+    Blog.all.each do |blog|
+      response += "<li>"
+      response += "<strong>Title: #{CGI.escape_html(blog.title)}</strong>, Content: #{CGI.escape_html(blog.content)}"
```
tomstuart: Kind of weird and annoying that we now have this mixture of manually escaping HTML ourselves (with `CGI.escape_html`) while also manually telling Action Controller _not_ to escape HTML (with `String#html_safe`). Fortunately we’ll deal with this in the next challenge. 😁

challenge_14/action_controller.rb
```diff
@@ -23,6 +24,44 @@ class Blog < ActiveRecord::Base; end
...
+class AppController < ActionController::Base
+  def root
+    response = ""
+    response += "<p><strong>Submit a new Blog Post!</p></strong>"
+    response += "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
+    response += "<p><label>Blog Title: <input name='title'></label></p>"
+    response += "<p><label>Content: <textarea name='content'></textarea></label></p>"
+    response += "<p><button>Submit post</button></p>"
+    response += "</form>"
+    render html: response.html_safe
```
tomstuart: I realise now that I misspoke when we were pairing: while `#html_safe` is indeed monkey-patched into `String`, it actually [returns an instance](https://github.com/rails/rails/blob/v6.1.3.1/activesupport/lib/active_support/core_ext/string/output_safety.rb#L312-L314) of `ActiveSupport::SafeBuffer`, not another `String` object as I claimed. (`SafeBuffer` is a subclass of `String` so this is only a small difference, but I didn’t want to mislead you!)

tomstuart (approved):
Well done! Great as usual. See below for miscellaneous “interesting“ remarks as usual. 😂

challenge_14/action_controller.rb
adrianna-chang-shopify: No we're not, sweeeet!!

challenge_14/action_controller.rb
adrianna-chang-shopify: Oh cool, I've definitely seen (and used!) `root` in `routes.rb` files in Rails apps, but I never fully pieced together that that's shorthand for `match '/', via: :get`!

challenge_14/action_controller.rb
adrianna-chang-shopify: Thank you for following up on this! ❤️

PR #14, opened 2021-05-05
Challenge 15 - Inline Action View Templates

adrianna-chang-shopify:
In this challenge, we render inline (`render :inline`) ERB templates in our `AppController` actions, rather than rendering strings with raw HTML (`render :html`).

ERB is a templating system that allows ruby code to be embedded in our HTML documents. ERB supplied to Action Controller's `render` method using the `:inline` option is automatically HTML-escaped (we no longer need to use `CGI.escape_html` to HTML escape our blog title / content).

The ERB template also has access to any instance variables defined in the controller action, allowing us to do this:
```ruby
    @blogs = Blog.all
    erb_response = <<~ERB
      <ul>
      <% @blogs.each do |blog| %>
        <li>
          <strong>Title: <%= blog.title %> </strong>, Content: <%= blog.content %>
        </li>
      <% end %>
      </ul>
    ERB
```

🔍  On Action View being separate from Action Pack
* Maybe so that Action View can be used in isolation from Action Controller / Action Dispatch
* Interestingly, Action Controller is still dependent on Action View, and [requires it as a dependency in `ActionController::Base`](https://github.com/rails/rails/blob/c049f16ae3731eeff9753bd479c828cda1e82b49/actionpack/lib/action_controller/base.rb#L3)
* However, this is still a one way dependency, so still sensible that Action View could be used in isolation

### DIFF
--- a/challenge_15/action_view.rb
+++ b/challenge_15/action_view.rb
@@ -1,0 +1,86 @@
+### Require dependencies
+require 'action_controller'
+require 'action_dispatch'
+require 'active_record'
+require 'cgi'
+
+class Blog < ActiveRecord::Base; end
+
+ActiveRecord::Base.establish_connection(
+  adapter:  "sqlite3",
+  database: "application.sqlite3"
+)
+
+# Create the table if it doesn't exist, seed some data
+ActiveRecord::Schema.define do
+  unless table_exists?(:blogs)
+    create_table :blogs do |t|
+      t.string :title
+      t.text :content
+    end
+    Blog.create!(title: 'My awesome blog!', content: 'my favourite HTML tags are <p> and <script>')
+    Blog.create!(title: 'Another cool blog!', content: 'my favourite HTML tags are <br> and <hr>')
+  end
+end
+
+class AppController < ActionController::Base
+  def root
+    response_body = ""
+    response_body += "<p><strong>Submit a new Blog Post!</p></strong>"
+    response_body += "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
+    response_body += "<p><label>Blog Title: <input name='title'></label></p>"
+    response_body += "<p><label>Content: <textarea name='content'></textarea></label></p>"
+    response_body += "<p><button>Submit post</button></p>"
+    response_body += "</form>"
+    render html: response_body.html_safe
+  end
+
+  def show_data
+    response_body = ""
+    response_body += "<ul>"
+
+    Blog.all.each do |blog|
+      response_body += "<li>"
+      response_body += "<strong>Title: #{CGI.escape_html(blog.title)}</strong>, Content: #{CGI.escape_html(blog.content)}"
+      response_body += "</li>"
+    end
+    response_body += "</ul>"
+    render html: response_body.html_safe
+  end
+
+  def create_post
+    puts 'Got a new POST request!'
+
+    # Alternatively, we can explicitly grab the params we want to pass to the #create method
+    # Blog.create!(title: params[:title], content: params[:content])
+    Blog.create!(params.permit(:title, :content))
+    redirect_to "/show-data", status: :see_other
+  end
+
+  def not_found
+    response_body = "Sorry, I don’t know what #{request.path_info} is"
+    render plain: response_body, status: :not_found
+  end
+end
+
+class MyApp
+  def initialize
+    @router = ActionDispatch::Routing::RouteSet.new
+    draw_routes
+  end
+
+  def call(environment)
+    @router.call(environment)
+  end
+
+  private
+
+  def draw_routes
+    @router.draw do
+      root to: AppController.action(:root)
+      get '/show-data', to: AppController.action(:show_data)
+      post 'create-post', to: AppController.action(:create_post)
+      match '*path', via: :all, to: AppController.action(:not_found)
+    end
+  end
+end
--- a/challenge_15/action_view.rb
+++ b/challenge_15/action_view.rb
@@ -2,7 +2,6 @@
 require 'action_controller'
 require 'action_dispatch'
 require 'active_record'
-require 'cgi'
 
 class Blog < ActiveRecord::Base; end
 
@@ -39,11 +38,13 @@ class AppController < ActionController::Base
   def show_data
     @blogs = Blog.all
 
+    # Note! ERB interpolated HTML is escaped unless marked as HTML safe / raw helper used
+
     erb_response = <<~ERB
       <ul>
       <% @blogs.each do |blog| %>
         <li>
-          <strong>Title: <%= CGI.escape_html(blog.title) %> </strong>, Content: <%= CGI.escape_html(blog.content) %>
+          <strong>Title: <%= blog.title %> </strong>, Content: <%= blog.content %>
         </li>
       <% end %>
       </ul>
--- a/challenge_15/action_view.rb
+++ b/challenge_15/action_view.rb
@@ -25,41 +25,47 @@ end
 
 class AppController < ActionController::Base
   def root
-    response_body = ""
-    response_body += "<p><strong>Submit a new Blog Post!</p></strong>"
-    response_body += "<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>"
-    response_body += "<p><label>Blog Title: <input name='title'></label></p>"
-    response_body += "<p><label>Content: <textarea name='content'></textarea></label></p>"
-    response_body += "<p><button>Submit post</button></p>"
-    response_body += "</form>"
-    render html: response_body.html_safe
+    erb_response = <<~ERB
+      <p><strong>Submit a new Blog Post!</p></strong>
+      <form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>
+      <label>Blog Title: <input name='title'></label></p>
+      <p><label>Content: <textarea name='content'></textarea></label></p>
+      <p><button>Submit post</button></p>
+      </form>
+    ERB
+    render inline: erb_response
   end
 
   def show_data
-    response_body = ""
-    response_body += "<ul>"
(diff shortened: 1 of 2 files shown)


challenge_15/action_view.rb
```diff
@@ -25,41 +25,47 @@ class Blog < ActiveRecord::Base; end
...
-    Blog.all.each do |blog|
-      response_body += "<li>"
-      response_body += "<strong>Title: #{CGI.escape_html(blog.title)}</strong>, Content: #{CGI.escape_html(blog.content)}"
-      response_body += "</li>"
-    end
-    response_body += "</ul>"
-    render html: response_body.html_safe
+    erb_response = <<~ERB
+      <ul>
+      <% @blogs.each do |blog| %>
```
tomstuart, line 45: As you noted, the controller’s instance variables are somehow magically visible from inside the template. Certainly easy and convenient — it’s so natural that most Rails developers never even think of it as a feature — but the underlying mechanism is actually kind of complicated!

It’s instructive to see how the pieces fit together:

* `ActionController::Base` [includes `ActionView::Layouts`](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_controller/base.rb#L213), which itself [includes `ActionView::Rendering`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/layouts.rb#L206), making all of the `ActionView::Rendering` methods available on a controller instance.
* `ActionController::Base` also [includes `AbstractController::Rendering`](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_controller/base.rb#L206), which provides the [main `#render` entry point](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/abstract_controller/rendering.rb#L23-L33) for template rendering. The primary responsibility of this method is to [call `#render_to_body`](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/abstract_controller/rendering.rb#L25) and [stick the resulting string in `self.response_body`](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/abstract_controller/rendering.rb#L32).
* The underlying `#render_to_body` method is available on the controller because it’s [implemented in `ActionView::Rendering`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/rendering.rb#L101-L104). All it really does it [call through](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/rendering.rb#L103) to a private method named [`#_render_template`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/rendering.rb#L108-L124).
* `#_render_template` [calls `#view_context`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/rendering.rb#L111) to construct the “view context”, which is ultimately the object that will function as the view, usually an `ActionView::Base` instance.
* `#view_context` [calls `#view_assigns`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/rendering.rb#L92) and passes it into the constructor of the view context class, usually `ActionView::Base`. Note that `#view_assigns` isn’t defined in `ActionView::Rendering` — it’s part of the module’s contract that we have to define `#view_assigns` ourselves if we mix `ActionView::Rendering` into our class.
* In our case, `#view_assigns` is implemented [back over in `AbstractController::Rendering`](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/abstract_controller/rendering.rb#L63-L69). You can see that it first [collects all of the controller’s instance variables](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/abstract_controller/rendering.rb#L64) (minus some protected ones) and then [puts their values into a hash](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/abstract_controller/rendering.rb#L67) keyed by variable name (i.e. with the leading `@` sliced off). Our `#view_assigns` will therefore contain `{ 'blogs' => Blog.all }` among other things.
* Remember how `#view_context` passed the `#view_assigns` hash into the [`ActionView::Base` constructor](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/base.rb#L230-L241)? Well, the constructor [calls `#assign`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/base.rb#L238) with that hash. `#assign` [iterates over the hash entries](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/base.rb#L211) and assigns each value to a correspondingly-named instance variable (after sticking the `@` back on, naturally). Our view context will therefore have `@blogs = Blog.all` at this point.
* So when the template [finally gets rendered](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/rendering.rb#L117) by [`ActionView::Renderer#render_to_object`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/renderer/renderer.rb#L25-L31), all of the controller’s instance variables are available on the view object and can be seen by the template (which [gets evaluated in the context of the view](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/template.rb#L300)), allowing us to say `@blogs.each` inside our ERB like it’s no big deal.

😅

challenge_15/action_view.rb
```diff
@@ -39,11 +38,13 @@ def root
...
     @blogs = Blog.all
 
+    # Note! ERB interpolated HTML is escaped unless marked as HTML safe / raw helper used
+
     erb_response = <<~ERB
       <ul>
       <% @blogs.each do |blog| %>
         <li>
-          <strong>Title: <%= CGI.escape_html(blog.title) %> </strong>, Content: <%= CGI.escape_html(blog.content) %>
+          <strong>Title: <%= blog.title %> </strong>, Content: <%= blog.content %>
```
tomstuart: The mechanism here is that [`ActionView::Template::Handlers::ERB::Erubi`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/template/handlers/erb/erubi.rb) is used as the [ERB implementation](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/template/handlers/erb.rb#L14) inside the [ERB template handler](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/template/handlers.rb#L14). This subclasses [`Erubi::Engine`](https://github.com/jeremyevans/erubi/blob/1.10.0/lib/erubi.rb#L47) from the [`erubi` gem](https://github.com/jeremyevans/erubi), and in particular [supplies an empty `ActionView::OutputBuffer`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/template/handlers/erb/erubi.rb#L32) for the template to be rendered into. `ActionView::OutputBuffer` [is a subclass](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/buffers.rb#L21) of [`ActiveSupport::SafeBuffer`](https://github.com/rails/rails/blob/v6.1.3.2/activesupport/lib/active_support/core_ext/string/output_safety.rb#L134), which [overrides string concatenation](https://github.com/rails/rails/blob/v6.1.3.2/activesupport/lib/active_support/core_ext/string/output_safety.rb#L186-L189) to [call `CGI.escapeHTML` on non-`html_safe?` strings](https://github.com/rails/rails/blob/v6.1.3.2/activesupport/lib/active_support/core_ext/string/output_safety.rb#L294-L296). So yeah, again, we get a feature so convenient that it’s essentially invisible — HTML escaping “just works” rather than us having to remember to call `CGI.escapeHTML` ourselves — only as a result of a complicated arrangement under the hood.

tomstuart (commented):
Not much to say here — it works, well done! For the sake of feeling useful I’ve added a little commentary on Rails’ inner workings for your entertainment/enlightenment (delete as appropriate).

tomstuart (approved):
Ahem. 😑

challenge_15/action_view.rb
adrianna-chang-shopify: WOW, this was quite the journey! 😂  Thanks for writing this up, I hadn't thought about unveiling the magic of how controller ivars are made available to the views 🤯  

Kind of random side question - what's the purpose of slicing off the `@` when storing the controller's ivars in a hash, if only to add the `@` back on when doing the instance variable set on the view context? Is there a reason we can't store `'@blogs'` directly in the hash?

challenge_15/action_view.rb
adrianna-chang-shopify: Very cool! Thanks for sharing ❤️

challenge_15/action_view.rb
tomstuart: Extremely good question! I don’t know the full answer to this, but it’s at least in part because `#render` supports [an `:assigns` option](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/rendering.rb#L110). If an `:assigns` hash is provided, it gets [passed into `#assign`](https://github.com/rails/rails/blob/v6.1.3.2/actionview/lib/action_view/rendering.rb#L113) just like the hash that comes out of `#view_assigns`, which results in additional instance variables being defined on the view context. So the style of keys in that hash is part of the public API of `#render`, and it’s aesthetically nicer to be able to say `assigns: { more_blogs: MoreBlog.all }` than `assigns: { '@more_blogs' => MoreBlog.all }` in your `render` call, hiding the messy detail that these values are going to end up as instance variables.

So in summary: by slicing off the `@` when automatically populating the `#view_assigns` hash, we make it look more like a hash that a human would write in their code, which makes it easier to incorporate actual human-written hashes into the same mechanism. If the hash was purely a private implementation detail then I don’t think there’d be any point in doing the back-and-forth conversion.

adrianna-chang-shopify: Ah neat!! Makes total sense to do this in the interest of providing a nicer API for `#render`. Thanks for the context! ❤️

PR #15, opened 2021-05-13
Challenges 16 + 17: Extract ERB templates and use tag helpers

adrianna-chang-shopify:
Nothing too fancy in this PR! We've extracted the inline ERB into template files, and told `ActionController` where to find them using:
```ruby
ActionController::Base.append_view_path('views')
```
ActionController's [default rendering behaviour](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-by-default-convention-over-configuration-in-action) is to look for a template in a directory that matches the controller name, with a filename that matches the action name.

So:
```ruby
class AppController < ActionController::Base
  def root
  end
```
will cause ActionController to look for `views/app/root.html.erb`.

💡 Fun Fact! You can omit an empty action method definition from a controller, and it ActionController will still render the associated template! (ie. we could omit `def root; end`). Rails ✨ magic ✨ 

We also introduced tag helpers, convenient ruby helper methods that build HTML compliant tags for us. While the simple ones like `tag.p` may not have a lot of added benefit in this case, the `form_tag` helper is a very useful helper because it handles many of the details for us, such as setting the default encoding type, defaulting to `/post` as the action, etc...

### DIFF
--- a/challenge_16/views/app/root.html.erb
+++ b/challenge_16/views/app/root.html.erb
@@ -14,5 +14,7 @@
       Content: <%= text_area_tag "content" %>
     <% end %>
   <% end %>
-  <%= button_tag "Submit post" %>
+  <%# #button_tag is an alternative option here, but #submit_tag is preferred when we're not %>
+  <%# embedding rich content / other tags within the button %>
+  <%= submit_tag "Submit post" %>
 <% end %>
--- a/challenge_16/views/app/root.html.erb
+++ b/challenge_16/views/app/root.html.erb
@@ -1,6 +1,18 @@
-<p><strong>Submit a new Blog Post!</p></strong>
-<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>
-<label>Blog Title: <input name='title'></label></p>
-<p><label>Content: <textarea name='content'></textarea></label></p>
-<p><button>Submit post</button></p>
-</form>
+<%= tag.p do %>
+  <%= tag.strong do %>
+    Submit a new Blog Post!
+  <% end %>
+<% end %>
+<%= form_tag("/create-post") do %>
+  <%= tag.p do %>
+    <%= label_tag do %>
+      Blog Title: <%= text_field_tag "title" %>
+    <% end %>
+  <% end %>
+  <%= tag.p do %>
+    <%= label_tag do %>
+      Content: <%= text_area_tag "content" %>
+    <% end %>
+  <% end %>
+  <%= button_tag "Submit post" %>
+<% end %>
--- a/challenge_16/views/app/root.html.erb
+++ b/challenge_16/views/app/root.html.erb
@@ -1,0 +1,6 @@
+<p><strong>Submit a new Blog Post!</p></strong>
+<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>
+<label>Blog Title: <input name='title'></label></p>
+<p><label>Content: <textarea name='content'></textarea></label></p>
+<p><button>Submit post</button></p>
+</form>
--- a/challenge_16/views/app/show_data.html.erb
+++ b/challenge_16/views/app/show_data.html.erb
@@ -1,7 +1,10 @@
-<ul>
-<% @blogs.each do |blog| %>
-  <li>
-    <strong>Title: <%= blog.title %> </strong>, Content: <%= blog.content %>
-  </li>
+<%= tag.ul do %>
+  <% @blogs.each do |blog| %>
+    <%= tag.li do %>
+      <%= tag.strong do %>
+        Title: <%= blog.title %>,
+      <% end %>
+      Content: <%= blog.content %>
+    <% end %>
+  <% end %>
 <% end %>
-</ul>
--- a/challenge_16/views/app/show_data.html.erb
+++ b/challenge_16/views/app/show_data.html.erb
@@ -1,9 +1,7 @@
 <%= tag.ul do %>
   <% @blogs.each do |blog| %>
     <%= tag.li do %>
-      <%= tag.strong do %>
-        Title: <%= blog.title %>,
-      <% end %>
+      <%= tag.strong do %>Title: <%= blog.title %><% end %>,
       Content: <%= blog.content %>
     <% end %>
   <% end %>
--- a/challenge_16/views/app/show_data.html.erb
+++ b/challenge_16/views/app/show_data.html.erb
@@ -1,0 +1,7 @@
+<ul>
+<% @blogs.each do |blog| %>
+  <li>
+    <strong>Title: <%= blog.title %> </strong>, Content: <%= blog.content %>
+  </li>
+<% end %>
+</ul>
--- a/challenge_16/config.ru
+++ b/challenge_16/config.ru
@@ -1,0 +1,17 @@
+require_relative 'final_project'
+
+class LoggerMiddleware
+  def initialize(app, string)
+    @app = app
+    @string = string
+  end
+
+  def call(environment)
+    puts @string
+    @app.call(environment)
+  end
+end
+
+use LoggerMiddleware, 'Rack app got a request!'
+
+run MyApp.new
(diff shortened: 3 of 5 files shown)


challenge_16/views/app/root.html.erb
```diff
@@ -1,6 +1,18 @@
-<p><strong>Submit a new Blog Post!</p></strong>
```
tomstuart: As we discussed: we actually had invalid HTML here! So one small benefit of the `tag` helper is that it takes care of closing tags in the right order for us. 😄

challenge_16/views/app/root.html.erb
```diff
@@ -1,6 +1,18 @@
...
-<form method='post' enctype='application/x-www-form-urlencoded' action='/create-post'>
-<label>Blog Title: <input name='title'></label></p>
```
tomstuart: And again! Guess we both missed the lack of an opening `<p>` tag here 😂

challenge_16/views/app/root.html.erb
```diff
@@ -1,6 +1,18 @@
...
+    <%= label_tag do %>
+      Blog Title: <%= text_field_tag "title" %>
+    <% end %>
+  <% end %>
+  <%= tag.p do %>
+    <%= label_tag do %>
+      Content: <%= text_area_tag "content" %>
+    <% end %>
+  <% end %>
+  <%= button_tag "Submit post" %>
```
tomstuart: As an HTML-related aside: we’re not getting any real benefit from using a `<button>` here instead of a more traditional `<input type="submit">`. As the [`#button_tag` documentation](https://api.rubyonrails.org/classes/ActionView/Helpers/FormTagHelper.html#method-i-button_tag) says:

>You can use the button tag as a regular submit tag but it isn't supported in legacy browsers. However, the button tag does allow for richer labels such as images and emphasis, so this helper will also accept a block.

So the purpose of `<button>` is really to be used with `type`s other than `submit` (i.e. `reset` or `button` — see [the `<button>` specification](https://html.spec.whatwg.org/#attr-button-type)), or with richer content than plain text (e.g. an image). We’re not doing either here, so `<input type="submit">` (i.e. `#submit_tag`) would be a more intention-revealing choice.

challenge_16/views/app/show_data.html.erb
```diff
@@ -1,7 +1,10 @@
-<ul>
-<% @blogs.each do |blog| %>
-  <li>
-    <strong>Title: <%= blog.title %> </strong>, Content: <%= blog.content %>
-  </li>
+<%= tag.ul do %>
+  <% @blogs.each do |blog| %>
+    <%= tag.li do %>
+      <%= tag.strong do %>
+        Title: <%= blog.title %>,
```
tomstuart: You’ve cheated a little here by moving the comma inside the `<strong>`! What gives?

challenge_16/views/app/show_data.html.erb
```diff
@@ -0,0 +1,7 @@
+<ul>
+<% @blogs.each do |blog| %>
```
tomstuart: This is not my favourite indentation style… (continued in comment on next commit)

challenge_16/views/app/show_data.html.erb
```diff
@@ -1,7 +1,10 @@
-<ul>
-<% @blogs.each do |blog| %>
-  <li>
-    <strong>Title: <%= blog.title %> </strong>, Content: <%= blog.content %>
-  </li>
+<%= tag.ul do %>
+  <% @blogs.each do |blog| %>
```
tomstuart, line 2: (continued from comment on previous commit) …but I do really appreciate this indentation style! I find ERB templates much easier to read when each tag increases the indentation level, regardless of whether it’s an HTML `<` or ERB `<%` tag. The downside is that the resulting HTML contains too much indentation once the ERB bits have been processed and removed, but since it’s very likely nobody will ever see the raw HTML anyway, I think it’s the right choice to prioritise clarity in the template.

tomstuart (approved):
Great job! I couldn’t really think of anything interesting to add this time 🤣

challenge_16/views/app/root.html.erb
adrianna-chang-shopify: Neat, that makes sense! I'll change this to a `#submit_tag` with a note 😁

challenge_16/views/app/show_data.html.erb
adrianna-chang-shopify: Ah yes, my bad! I noticed that moving it to be outside the block to `tag.strong` was introducing an extra space:
```ruby
    <%= tag.li do %>
      <%= tag.strong do %>
        Title: <%= blog.title %>
      <% end %>,
      Content: <%= blog.content %>
    <% end %>
```
Screen Shot 2021-05-14 at 8 58 51 AM

So I moved it into the block to experiment and then forgot about it 😅 

Putting the block on a single line with the tag helper fixes this:
```ruby
<%= tag.strong do %>Title: <%= blog.title %><% end %>,
```
Screen Shot 2021-05-14 at 9 01 51 AM

Is it bad style to have a `do; end` block across one line, or would this be the best way to do it?

challenge_16/views/app/show_data.html.erb
adrianna-chang-shopify: Interesting, I never realized this was an "improper" way to write these ERB templates (ie. that the resulting raw HTML wouldn't be considered "correct")! 

Can you just clarify what the correct way to write this would be to have the correct amount of indentation? Is it only `<%` tags that should be indented?

challenge_16/views/app/root.html.erb
adrianna-chang-shopify: Wow, good thing I don't write HTML for a living 😂

challenge_16/views/app/show_data.html.erb
tomstuart: I fear there’s no perfect solution here. 😢 

Whitespace in ERB is a bit of a pain. The [`ActionView::Base` documentation](https://api.rubyonrails.org/classes/ActionView/Base.html#class-ActionView::Base-label-ERB) says:

>When on a line that only contains whitespaces except for the tag, `<% %>` suppresses leading and trailing whitespace, including the trailing newline. `<% %>` and `<%- -%>` are the same. Note however that `<%= %>` and `<%= -%>` are different: only the latter removes trailing whitespaces.

I think this behaviour comes from [erubi](https://github.com/jeremyevans/erubi) but, either way, it’s neither very reliable nor very well documented. In any case, it does reveal that ERB has some whitespace-suppression features which can be turned on by adding the `-` character inside the tags. So I think this would do what you want…

```erb
<%= tag.ul do %>
  <% @blogs.each do |blog| %>
    <%= tag.li do %>
      <%= tag.strong do %>
        Title: <%= blog.title -%>
<% end %>,
      Content: <%= blog.content %>
    <% end %>
  <% end %>
<% end %>
```

…by suppressing the newline at the end of the `Title:` line (and manually removing the indentation before the following comma), but that’s not very nice!

Your solution of putting the whole `<strong>` tag on one line, thereby avoiding any newlines or indentation inside the tag altogether, is probably the least bad option. I wouldn’t say it’s bad style; it’s not ideal but it’s sometimes necessary for this exact reason.

challenge_16/views/app/show_data.html.erb
tomstuart: There’s no single correct way to do this, so it’s just a question of taste. If you wanted the HTML to have beautiful indentation for whatever reason, then you’d need to ignore the `<%` tags from the point of view of indentation (i.e. adding `<% … %>` … `<% end %>` tags around a section of HTML would not affect its indentation level), so each HTML tag ends up indented by the apparently-correct amount once all the ERB tags are removed during rendering. But to be completely clear, I don’t think anyone except me actually cares either way!

challenge_16/views/app/root.html.erb
tomstuart: Likewise!

challenge_16/views/app/show_data.html.erb
adrianna-chang-shopify: Gotcha! Thanks for digging into the details for me ❤️ In writing ERB, it does often feel like you have to "cheat" a bit in one way or another to get exactly what you want 🤪

PR #16, opened 2021-05-27
Challenges 18 + 19 - Shorthand routes & resourceful routes!

adrianna-chang-shopify:
https://github.com/adrianna-chang-shopify/learning-project/commit/c6667a22977fc1297a5b13e374c38993ceafbaaa - introduces [shorthand routes](https://www.rubydoc.info/docs/rails/ActionDispatch%2FRouting%2FMapper%2FBase:match). Instead of using `AppController.action(:root)` we can use `app#root` as a shorthand (ie. `controller_name#action_name`).

https://github.com/adrianna-chang-shopify/learning-project/commit/ce3d4d730d33478e3e4a73dac4702650d70ec163, https://github.com/adrianna-chang-shopify/learning-project/commit/06f7d274ecdc48cb4892f850ec8dcba746547351, https://github.com/adrianna-chang-shopify/learning-project/commit/f41de305251f288ed46a909d260fdd0b931ba548 use an incremental approach to getting our application to use resourceful routes:
- First, we rename our controller to match the name of our `ActiveRecord::Relation` (`BlogsController`)
- Next, we use the conventional Rails CRUD actions in our controller (`new`, `index` and `create`) and update our request paths to match the conventions as defined [here](https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions)
- Finally, we switch our custom routes our for resourceful routes: `resources :blogs`

https://github.com/adrianna-chang-shopify/learning-project/commit/a0057c8bde9d21e12ad3abb42eaf41993009e690 extracts the `:not_found` controller action from the `BlogsController` to a generic `PagesController` so that our `BlogsController` only handles CRUD actions related to `Blogs`

Finally, in https://github.com/adrianna-chang-shopify/learning-project/commit/607d968fe435b329e8035fc4657d510c9af10c9c we refactor our application such that it's no longer wrapped in a Ruby class. Previously, having a Ruby class for `MyApp` allowed us to organize drawing our routes a bit better, and we were storing the `Ad::Routing::RouteSet` in an ivar that we could call `.call` on.

Now, we instantiate the routeset directly, assign it to the `MyApp` constant, and allow it to be `.call`ed implicitly from `config.ru` when we do `run MyApp`.


adrianna-chang-shopify:
cc @tomstuart in case you missed this! 😁

tomstuart (approved):
Oops, sorry I forgot about this! I did check through it once last week and couldn’t think of anything clever to add, so I hoped if I waited I’d think of something clever, but… bad news, I didn’t. Oh well. This is all excellent, well done 🎉 😂

PR #17, opened 2021-06-04
🚀 😱 🤯  FINAL CHALLENGE - Use URL helpers

adrianna-chang-shopify:
In this PR, we introduce path / URL helpers made available to us via an (anonymous) module returned via [`ActionDispatch::Routing::RouteSet#url_helpers`](https://github.com/rails/rails/blob/v6.1.3/actionpack/lib/action_dispatch/routing/route_set.rb#L479-L485). All we have to do is include this module in our controller, ie. `BlogsController.include(MyApp.url_helpers)` and voila! We have access to these handy dandy helpers ✨ 

I had a couple questions @tomstuart !
1) In the doc, when you say
> We can either call the helper methods on this module object directly

Is this at all possible with our setup, or is the only way to access these helpers to include the module in the controller when we instantiate the `RouteSet`? Since we don't have a full Rails application I realize we can't do `Rails.application.routes.url_helpers`, but is there some other way for our views / controllers to access the routeset that you know of? 🤔 

2) I'm curious on the distinction between path and URL helpers. From the (light) research I did, many seemed to suggest that URL helpers are important if the URL will be consumed externally to the web app (ie. email) but also if a redirect is being used. For the latter, many claimed it to be because the HTTP specification indicates that an absolute URL is required for the `Location` header in any 3xx responses:
> For 3xx responses, the location SHOULD indicate the server's preferred URI for automatic redirection to the resource. The field value consists of a single absolute URI.
 Location       = "Location" ":" absoluteURI

According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.30

However, the RFC we were originally looking at [says this:](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.2)

>      Location = URI-reference

> The field value consists of a single URI-reference.  When it has the
   form of a relative reference ([RFC3986], Section 4.2), the final
   value is computed by resolving it against the effective request URI
   ([RFC3986], Section 5).

>  If the Location value provided in a 3xx (Redirection) response does
   not have a fragment component, a user agent MUST process the
   redirection as if the value inherits the fragment component of the
   URI reference used to generate the request target (i.e., the
   redirection inherits the original reference's fragment, if any).

I'm kind of confused as to whether this is saying a different thing (that relative URLs are okay..?) and what the implications are for the redirect in our create action:
```ruby
  def create
    Blog.create!(params.permit(:title, :content))
    redirect_to blogs_url, status: :see_other
  end
```

Can you offer any wisdom here? 😄

### DIFF
--- a/last_challenge/final_project.rb
+++ b/last_challenge/final_project.rb
@@ -36,7 +36,7 @@ class BlogsController < ActionController::Base
   def create
     puts 'Got a new POST request!'
     Blog.create!(params.permit(:title, :content))
-    redirect_to "/blogs", status: :see_other
+    redirect_to blogs_url, status: :see_other
   end
 end
 
@@ -49,6 +49,7 @@ end
 
 # Top level entry point to Rack application is the RouteSet!
 MyApp = ActionDispatch::Routing::RouteSet.new
+BlogsController.include(MyApp.url_helpers)
 MyApp.draw do
   resources :blogs
   match '*path', via: :all, to: 'pages#not_found'
--- a/last_challenge/final_project.rb
+++ b/last_challenge/final_project.rb
@@ -36,7 +36,9 @@ class BlogsController < ActionController::Base
   def create
     puts 'Got a new POST request!'
     Blog.create!(params.permit(:title, :content))
-    redirect_to blogs_url, status: :see_other
+    # Since we're accessing the URL helpers right on the module, we need to specify
+    # both the host and port so that the full (absolute) URL can be constructed properly
+    redirect_to MyApp.url_helpers.blogs_url(host: request.host, port: request.port), status: :see_other
   end
 end
 
@@ -49,7 +51,10 @@ end
 
 # Top level entry point to Rack application is the RouteSet!
 MyApp = ActionDispatch::Routing::RouteSet.new
-BlogsController.include(MyApp.url_helpers)
+# Can mix the module with URL helpers right into the controller,
+# or access them as singleton methods on the module returned by
+# RouteSet#url_helpers
+# BlogsController.include(MyApp.url_helpers)
 MyApp.draw do
   resources :blogs
   match '*path', via: :all, to: 'pages#not_found'
--- a/last_challenge/views/blogs/new.html.erb
+++ b/last_challenge/views/blogs/new.html.erb
@@ -3,7 +3,7 @@
     Submit a new Blog Post!
   <% end %>
 <% end %>
-<%= form_tag("/blogs") do %>
+<%= form_tag(blogs_path) do %>
   <%= tag.p do %>
     <%= label_tag do %>
       Blog Title: <%= text_field_tag "title" %>
--- a/last_challenge/views/blogs/new.html.erb
+++ b/last_challenge/views/blogs/new.html.erb
@@ -3,7 +3,7 @@
     Submit a new Blog Post!
   <% end %>
 <% end %>
-<%= form_tag(blogs_path) do %>
+<%= form_tag(MyApp.url_helpers.blogs_path) do %>
   <%= tag.p do %>
     <%= label_tag do %>
       Blog Title: <%= text_field_tag "title" %>


last_challenge/final_project.rb
```diff
@@ -36,7 +36,9 @@ def index
   def create
     puts 'Got a new POST request!'
     Blog.create!(params.permit(:title, :content))
-    redirect_to blogs_url, status: :see_other
+    # Since we're accessing the URL helpers right on the module, we need to specify
+    # both the host and port so that the full (absolute) URL can be constructed properly
+    redirect_to MyApp.url_helpers.blogs_url(host: request.host, port: request.port), status: :see_other
```
tomstuart: During pairing we puzzled over how the mixed-in version of `#blogs_url` works: does the URL helper have some privileged access to the controller which allows it to infer `host:` and `port:` without them being provided explicitly?

I have investigated! Here’s what happens.

The [anonymous module](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_dispatch/routing/route_set.rb#L490) returned by `ActionDispatch::Routing::RouteSet#generate_url_helpers` both [extends `ActiveSupport::Concern`](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_dispatch/routing/route_set.rb#L491) and [includes `ActionDispatch::Routing::UrlFor`](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_dispatch/routing/route_set.rb#L492).

The former is relevant because, unlike Ruby’s built-in [`Module.included` hook](https://rubydoc.info/stdlib/core/2.7.0/Module:included), it provides the ability for a module to [inherit hooks](https://api.rubyonrails.org/v6.1.3.2/classes/ActiveSupport/Concern.html) from its dependencies:

```
>> require 'active_support/concern'
=> true
>> module A
     extend ActiveSupport::Concern
     included do |base|
       puts "included A in #{base}"
     end
   end
=> #<Proc>
>> module B
     extend ActiveSupport::Concern
     include A
   end
=> B
>> module C
     include B
   end
included A in C
=> C
```

The latter is relevant because `ActionDispatch::Routing::UrlFor` defines [an `included` hook](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_dispatch/routing/url_for.rb#L91-L104) which sets up an Active Support “class attribute” called `default_url_options` (either [inheritable by subclasses](https://api.rubyonrails.org/classes/Class.html#method-i-class_attribute) or [not](https://api.rubyonrails.org/classes/Module.html#method-i-mattr_writer), depending on whether it’s being set up in a class or a module). So, because `UrlFor` [is itself](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_dispatch/routing/url_for.rb#L88) an `ActiveSupport::Concern`, this hook gets inherited by the anonymous URL helper module and therefore runs against whichever class that module eventually gets mixed into (e.g. our controller) and sets up a `default_url_options` class attribute there.

These Active Support class attributes are a little unusual in that they’re available at both the class and instance level:

```
>> require 'active_support/core_ext/class'
=> true
>> class MyClass
     class_attribute :my_attribute
   end
=> [:my_attribute]
>> MyClass.my_attribute = 'hello'
=> "hello"
>> my_instance = MyClass.new
=> #<MyClass>
>> my_instance.my_attribute
=> "hello"
```

This is used by the `UrlFor#url_options` instance method to [default to](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_dispatch/routing/url_for.rb#L115-L117) the value of the `default_url_options` class attribute; the [URL generation code](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_dispatch/routing/url_for.rb#L176-L181) itself fetches options via this `#url_options` wrapper rather than directly from `default_url_options`.

Action Controller knows about this indirection and takes advantage of it by [overriding](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_controller/metal/url_for.rb#L30) `#url_options` to [merge information from the current request](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_controller/metal/url_for.rb#L31-L36) into the `default_url_options` it gets from `super`. So each controller instance gets the values of `request.host` and `request.optional_port` frozen into its `@_url_options` instance variable the first time `#url_options` is called, and all of the calls to URL helpers within the course of a request get these values supplied implicitly.

Which is a long way of saying that URL helpers always call `#url_options` on the receiver, and you can supply implicit options by overriding it! So something like this works:

```
>> require 'action_dispatch'
=> true
>> MyApp = ActionDispatch::Routing::RouteSet.new
=> #<ActionDispatch::Routing::RouteSet>
>> MyApp.draw do
     resources :blogs
     match '*path', via: :all, to: 'pages#not_found'
   end
=> nil
>> class RackEndpoint
     include MyApp.url_helpers
     attr_accessor :url_options
     
     def call(env)
       self.url_options = { host: env['SERVER_HOST'], port: env['SERVER_PORT'] }
       [303, { 'Location' => blogs_url({}) }, []]
     end
   end
=> :call
>> RackEndpoint.new.call('SERVER_HOST' => 'localhost', 'SERVER_PORT' => 1234)
=> [303, {"Location"=>"http://localhost:1234/blogs"}, []]
```

Of course, if we (somehow) don’t need `:host` and `:port` to be dynamic per-request, we can just have `#url_options` always return the same thing instead of squirrelling request details away in the instance’s state:

```
>> class RackEndpoint
     include MyApp.url_helpers
     
     def url_options
       { host: 'localhost', port: '1234' }
     end
     
     def call(env)
       [303, { 'Location' => blogs_url({}) }, []]
     end
   end
=> :call
>> RackEndpoint.new.call({})
=> [303, {"Location"=>"http://localhost:1234/blogs"}, []]
```

And finally: the anonymous URL helper module itself complies with this mechanism by defining a [dummy `.url_options` class method](https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_dispatch/routing/route_set.rb#L540), so when we call a helper on the module directly, e.g. `MyApp.url_helpers.blogs_url`, it’s able to successfully fetch an empty hash from `MyApp.url_helpers.url_options` rather than blowing up.

last_challenge/final_project.rb
```diff
@@ -36,7 +36,7 @@ def index
   def create
     puts 'Got a new POST request!'
     Blog.create!(params.permit(:title, :content))
-    redirect_to "/blogs", status: :see_other
+    redirect_to blogs_url, status: :see_other
```
tomstuart: > I'm kind of confused as to whether [RFC 7231] is saying a different thing [to RFC 2616] (that relative URLs are okay..?) and what the implications are for the redirect in our create action.
>
> Can you offer any wisdom here? 😄

Well I dunno about _wisdom_, but I can offer some history as a compromise 😂 

To understand what’s happened we need to take a look at the history of the HTTP specification! The [W3C’s page](https://www.w3.org/Protocols/History.html) on this topic starts at [RFC 1945](https://datatracker.ietf.org/doc/html/rfc1945) (HTTP/1.0) in 1996 and ends at [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616) (HTTP/1.1) in 1999.

For this entire period the advice on the `Location` header remains basically unchanged. RFC 1945 [says](https://datatracker.ietf.org/doc/html/rfc1945#section-10.11):

> The Location response-header field defines the exact location of the resource that was identified by the Request-URI. For 3xx responses, the location must indicate the server's preferred URL for automatic redirection to the resource. Only one absolute URL is allowed.

Whereas RFC 2616 [says](https://datatracker.ietf.org/doc/html/rfc2616#section-14.30):

> The Location response-header field is used to redirect the recipient to a location other than the Request-URI for completion of the request or identification of a new resource. For 201 (Created) responses, the Location is that of the new resource which was created by the request. For 3xx responses, the location SHOULD indicate the server's preferred URI for automatic redirection to the resource. The field value consists of a single absolute URI.

Which is a little more detailed, and uses more consistently precise language (the technically correct “[URI](https://datatracker.ietf.org/doc/html/rfc2396#section-1)” rather than the more common “[URL](https://datatracker.ietf.org/doc/html/rfc1738)”), but is otherwise saying essentially the same thing: the `Location` header contains an absolute URL that tells the client where to find the resource it’s looking for.

But beyond this point we can start to see the work of [httpbis](https://datatracker.ietf.org/wg/httpbis/charter/), an IETF working group responsible for refining, clarifying and expanding the HTTP specification (“bis” = “[repeated](https://en.wiktionary.org/wiki/bis#Etymology_2)”). One of their decisions in 2007 was to begin splitting up the monolithic HTTP/1.1 specification into several more focused documents, namely:

* [RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230) — Message Syntax and Routing
* [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231) — Semantics and Content
* [RFC 7232](https://datatracker.ietf.org/doc/html/rfc7232) — Conditional Requests
* [RFC 7233](https://datatracker.ietf.org/doc/html/rfc7233) — Range Requests
* [RFC 7234](https://datatracker.ietf.org/doc/html/rfc7234) — Caching
* [RFC 7235](https://datatracker.ietf.org/doc/html/rfc7235) — Authentication

(This split is recorded at the top of [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616), where it now says “Obsoleted by: 7230, 7231, 7232, 7233, 7234, 7235”.)

[RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231), “Semantics and Content”, is where the relevant section ended up. As you pointed out, it now [says](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.2) something different:

> The field value consists of a single URI-reference.  When it has the form of a relative reference ([RFC3986], Section 4.2), the final value is computed by resolving it against the effective request URI ([RFC3986], Section 5).
> […]
> If the Location value provided in a 3xx (Redirection) response does not have a fragment component, a user agent MUST process the redirection as if the value inherits the fragment component of the URI reference used to generate the request target (i.e., the redirection inherits the original reference's fragment, if any).

So there’s some nontrivial change here! Apparently the URL can now be a [relative reference](https://datatracker.ietf.org/doc/html/rfc3986#section-4.2) and can contain a [fragment identifier](https://datatracker.ietf.org/doc/html/rfc3986#section-3.5).

By clicking on the “From: [draft-ietf-httpbis-p2-semantics-26](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-p2-semantics-26)” link at the top of RFC 7231 we can investigate when these changes happened, because the header of the draft contains links to the history of all its revisions.

Flipping through the history, we can see that:
* In [version 1](https://www.ietf.org/rfcdiff?url2=draft-ietf-httpbis-p2-semantics-01.txt#part-14) of the draft (January 2008) they add the ability to include a fragment identifier at the end of the URL. I expect this is merely reflecting something that was already happening in real implementations, so they took this opportunity to call it out explicitly and enumerate the circumstances in which it’s not appropriate to do it.
* In [version 8](https://www.ietf.org/rfcdiff?url2=draft-ietf-httpbis-p2-semantics-08.txt#part-18) (October 2009) they remove the language about it needing to explicitly be an _absolute_ URL. Again I expect this was acknowledging the reality, i.e. that some HTTP servers were already sending relative URLs in `Location` headers and that therefore HTTP clients were already having to deal with it, without yet getting into the details of how relative URLs should be handled. Notice that relaxing this constraint allows them to switch from `absolute-URI` to `URI` in their definition of the header syntax and thus piggy-back off the [`URI` syntax definition](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A) in RFC 3986, which already covers the ability to include a fragment identifier, so this is a nice simplifying unification of related standards.
* In [version 10](https://www.ietf.org/rfcdiff?url2=draft-ietf-httpbis-p2-semantics-10.txt#part-23) (July 2010) they actually do get into the details of how relative URLs should be handled. In particular they lean on [section 5](https://datatracker.ietf.org/doc/html/rfc3986#section-5) of RFC 3986, which is all about how relative URLs should be resolved relative to some context, in this case the URL of the original request. Again this harmonises the HTTP and URI specifications by allowing one to defer to the official word of the other rather than inventing some similar-but-different mechanism. They also have to add a note at this point to highlight the fact that it’s unclear (and therefore unspecified) how fragment identifiers should interact when they appear in both the original request URL and a relative `Location` URL — overall they’ve given this header two new features, but we should avoid using both of them at once.
* In [version 18](https://www.ietf.org/rfcdiff?url1=draft-ietf-httpbis-p2-semantics-18.txt#part-27) (March 2012) they eventually relent and make a decision about how to handle conflicting fragment identifiers in this particular case.

So… I’m sorry that I don’t have real knowledge about the reasons for the decisions above, but I hope that this gives you some flavour of the way these specifications evolve over time and how the committees involved try to capture more real-world subtleties and edge cases as time goes on. And I guess the tldr is: yes, the spec has changed and relative URLs are now okay!

tomstuart (approved):
A small but perfectly-formed change, well done. I’ve tried to answer your (express and implied) questions in comments.

Congratulations on finishing this project! 🎉 👏🏻

last_challenge/final_project.rb
adrianna-chang-shopify: This is awesome! It's really cool that you can see the diffs for the different versions to compare exactly what changed. It makes a lot of sense that these specs weren't written in their final form the first time around 😄  Really neat, thanks for sharing! ❤️

last_challenge/final_project.rb
adrianna-chang-shopify: This writeup was epic and immensely valuable - thank you!!

